PR #33, opened 2024-03-19
[Fix] Various fixes such that this can be built using ament

floriantschopp:
# Description
* Fully moving to cpp20, removing all `std::format`
* Some missing includes
* Relax required cmake version
* Remove an annoying false positive check for `clang-tidy`

## Type of change


filippobrizzi (approved):
Thank you!

PR #34, opened 2024-03-22
fix: Add the ability to serialize templated types.

lhruby:
# Description
To allow serialization of templated types, we add variadic template parameters to `struct ProtoAssociation`.

## Type of change

### DIFF
--- a/modules/serdes/tests/tests.cpp
+++ b/modules/serdes/tests/tests.cpp
@@ -2,6 +2,8 @@
 // Copyright (C) 2023-2024 HEPHAESTUS Contributors
 //=================================================================================================
 
+#include <cstddef>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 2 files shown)


modules/serdes/tests/tests.cpp
```diff
@@ -31,13 +33,28 @@ struct Data {
...
 }  // namespace heph::serdes::tests
 
 namespace heph::serdes::protobuf {
 template <>
 struct ProtoAssociation<heph::serdes::tests::Data> {
   using Type = heph::serdes::tests::MockProtoMessage;
 };
+
+template <class... Args>
+    struct ProtoAssociation < heph::serdes::tests::TemplatedData<Args...> {
```
filippobrizzi, line 55: you are missing a closing `>`

lhruby:
```
void fromProto(const MockProtoMessage& proto, Data& data) {
  (void)proto;
  data.a = NUMBER * 2;
}
```
How is serialization tested? As is, it seems like the return is fixed.

filippobrizzi:
> ```
> void fromProto(const MockProtoMessage& proto, Data& data) {
>   (void)proto;
>   data.a = NUMBER * 2;
> }
> ```
> 
> How is serialization tested? As is, it seems like the return is fixed.

This module just provides high-level abstract function. If you want to test with real messages I guess there are two options.
* Faster: Add a new type in examples and add a test using the feature there
* Better: Create a proto folder inside `test` and add there a proto message that checks your change

PR #35, opened 2024-03-22
Add spinner and concurrency module

floriantschopp:
# Description
Added a spinner that can be used for running applications using the ipc module.

## Type of change


modules/concurrency/src/spinner.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+Spinner::~Spinner() {
+  try {
+    if (is_started_.load() && spinner_thread_.joinable()) {
+      stop();
+    }
+  } catch (const std::exception& ex) {
+    fmt::print("Error stopping spinner: {}", ex.what());
+  }
```
filippobrizzi, line 22: I am wondering whether here we should report an error if stop is not called before the destructor.
Maybe even throw, even if it is not recommended inside a constructor.

Do you remember the start/stop behavior in Arrival?

modules/concurrency/src/spinner.cpp
```diff
@@ -0,0 +1,56 @@
...
+  throwExceptionIf<InvalidOperationException>(is_started_.load(), fmt::format("Spinner is already started."));
+
+  spinner_thread_ = std::jthread([this](std::stop_token stop_token) { spin(stop_token); });
+
+  is_started_.store(true);
+}
+
+void Spinner::spin(std::stop_token& stop_token) {
+  while (!stop_token.stop_requested()) {
+    spinOnce();
```
filippobrizzi, line 35: Do you remember in arrival that we had issue of exception not being reported properly, was that because of std::async right? here we should be good I think

modules/concurrency/src/spinner.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+  if (stop_callback_) {
+    stop_callback_();
+  }
+
+  is_started_.store(false);
+}
+
+void Spinner::addStopCallback(std::function<void()> callback) {
+  stop_callback_ = std::move(callback);
```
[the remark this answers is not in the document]
floriantschopp: if you don't want to override the `stop` function but still want some specific behavour to happen when the spinner stops, e.g. write some file, (save a map), ...

modules/concurrency/src/spinner.cpp
floriantschopp: yes, I think it was because of the async.

modules/concurrency/src/spinner.cpp
floriantschopp: this code is actualy to not throw in the destructor, as it won't be  propagated correctly. Why do you think its not good design here?

floriantschopp:
@filippobrizzi I can't seem to get `clang-format` happy :(

modules/concurrency/src/spinner.cpp
filippobrizzi: It should be an error if the user doesn't call stop before the spinner terminates. So I think we should be destructive (pun intended).
Maybe I would even do `std::terminate`.

modules/concurrency/src/spinner.cpp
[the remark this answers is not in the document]
filippobrizzi: ah nice!

filippobrizzi (approved):

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -0,0 +1,34 @@
...
+
+  void start();
+  void stop();
+  virtual void spin();
+  virtual void spinOnce() = 0;  // Pure virtual function
+  void addStopCallback(std::function<void()> callback);
+
+private:
+  std::atomic<bool> is_started_ = false;
+  std::atomic<bool> stop_requested_ = false;
```
filippobrizzi, line 30: cosmetics:
I don't know if like this is more clean, WDYT?
```suggestion
  std::atomic_bool is_started_ = false;
  std::atomic_bool stop_requested_ = false;
```

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -0,0 +1,34 @@
...
+  Spinner();
+  virtual ~Spinner();
+  Spinner(const Spinner&) = delete;
+  auto operator=(const Spinner&) -> Spinner& = delete;
+  Spinner(Spinner&&) = delete;
+  auto operator=(Spinner&&) -> Spinner& = delete;
+
+  void start();
+  void stop();
+  virtual void spin();
```
filippobrizzi, line 24: I think this should be private and non-virtual. 
If we want to add more logic we can do it and have a configurator that controls it.
For example, we could have 
```
struct SpinnerParams {
 std::chrono::duration period;
}
// ...

void spin() {
  while(!stop_requested_) {
    spinOnce();
    if (period > 0) { sleep_until(...); }|
  }
}
}
```
but in general I would say that we shouldn't let the user the option to control this

modules/concurrency/src/spinner.cpp
```diff
@@ -0,0 +1,55 @@
...
+  is_started_.store(true);
+}
+
+void Spinner::spin() {
+  while (!stop_requested_.load()) {
+    spinOnce();
+  }
+}
+
+void Spinner::stop() {
```
filippobrizzi, line 38: given that this can take time to terminate maybe we want to return a `std::future`?

modules/concurrency/include/hephaestus/concurrency/spinner.h
floriantschopp: agree, looks better!

modules/concurrency/include/hephaestus/concurrency/spinner.h
floriantschopp: fair enough, lets for now just make it private, if we need more control we can do once its needed.

modules/concurrency/src/spinner.cpp
floriantschopp: do we need this to be async? can be blocking, no?

filippobrizzi: it could be nice if we control multiple spinner we can call stop in parallel rather than one by one and waiting each one to terminate

floriantschopp: done

PR #40, opened 2024-04-03
Fix for gcc compilation

filippobrizzi:
# Description
Some fixes to make hephaestus compile with GCC

* GCC doesn't support `constexpr` strings
* GCC doesn't care if we cover all switch cases for an enum
* GCC think some variable are shadowed

## Type of change

### DIFF
--- a/modules/cli/include/hephaestus/cli/program_options.h
+++ b/modules/cli/include/hephaestus/cli/program_options.h
@@ -69,7 +69,7 @@ class ProgramDescription {
 public:
   /// @brief Creates object
   /// @param brief A brief text describing the program
-  constexpr explicit ProgramDescription(std::string brief);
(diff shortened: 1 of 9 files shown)


modules/cli/include/hephaestus/cli/program_options.h
```diff
@@ -69,15 +69,15 @@ class ProgramDescription {
 public:
   /// @brief Creates object
   /// @param brief A brief text describing the program
-  constexpr explicit ProgramDescription(std::string brief);
+  explicit ProgramDescription(std::string brief);  // NOTE: GCC doesn't like this to be constexpr
```
floriantschopp, line 72: probably remove the note here or add it to all below

floriantschopp (approved):

lhruby (approved):
LGTM.

PR #44, opened 2024-04-05
File utils

filippobrizzi:
# Description
Add utils to read and write to files

### DIFF
--- a/modules/utils/include/hephaestus/utils/file.h
+++ b/modules/utils/include/hephaestus/utils/file.h
@@ -21,9 +21,9 @@ public:
   auto operator=(const SelfDestructingPath&) -> SelfDestructingPath& = delete;
   auto operator=(SelfDestructingPath&&) -> SelfDestructingPath& = default;
 
-  [[nodiscard]] static auto tempFile() -> SelfDestructingPath;
+  [[nodiscard]] static auto createFile() -> SelfDestructingPath;
 
-  [[nodiscard]] static auto tempDir() -> SelfDestructingPath;
+  [[nodiscard]] static auto createDir() -> SelfDestructingPath;
 
   operator std::filesystem::path() const;  // NOLINT(google-explicit-conversion,
                                            // google-explicit-constructor)
--- a/modules/utils/include/hephaestus/utils/file.h
+++ b/modules/utils/include/hephaestus/utils/file.h
@@ -12,18 +12,18 @@ void writeBufferToFile(const std::filesystem::path& path, std::span<const std::b
 
 /// This class allows to create a file that is removed when the class goes out of scope (RAII).
 /// This is very useful in tests to avoid having dangling files.
-class SelfDestructingPath {
+class ScopedFilesystemPath {
 public:
-  explicit SelfDestructingPath(std::filesystem::path path);
-  ~SelfDestructingPath();
-  SelfDestructingPath(const SelfDestructingPath&) = delete;
-  SelfDestructingPath(SelfDestructingPath&&) = default;
-  auto operator=(const SelfDestructingPath&) -> SelfDestructingPath& = delete;
-  auto operator=(SelfDestructingPath&&) -> SelfDestructingPath& = default;
+  explicit ScopedFilesystemPath(std::filesystem::path path);
+  ~ScopedFilesystemPath();
+  ScopedFilesystemPath(const ScopedFilesystemPath&) = delete;
+  ScopedFilesystemPath(ScopedFilesystemPath&&) = default;
+  auto operator=(const ScopedFilesystemPath&) -> ScopedFilesystemPath& = delete;
+  auto operator=(ScopedFilesystemPath&&) -> ScopedFilesystemPath& = default;
 
-  [[nodiscard]] static auto createFile() -> SelfDestructingPath;
+  [[nodiscard]] static auto createFile() -> ScopedFilesystemPath;
 
-  [[nodiscard]] static auto createDir() -> SelfDestructingPath;
+  [[nodiscard]] static auto createDir() -> ScopedFilesystemPath;
 
   operator std::filesystem::path() const;  // NOLINT(google-explicit-conversion,
                                            // google-explicit-constructor)
--- a/modules/utils/include/hephaestus/utils/file.h
+++ b/modules/utils/include/hephaestus/utils/file.h
@@ -1,0 +1,44 @@
+#pragma once
+#include <filesystem>
+#include <span>
+
+namespace heph::utils {
+
+[[nodiscard]] auto readFile(const std::filesystem::path& path) -> std::string;
+[[nodiscard]] auto readBinaryFile(const std::filesystem::path& path) -> std::vector<std::byte>;
+
+void writeStringToFile(const std::filesystem::path& path, std::string_view content);
+void writeBufferToFile(const std::filesystem::path& path, std::span<std::byte> content);
+
+/// This class allows to create a file that is removed when the class goes out of scope (RAII).
+/// This is very useful in tests to avoid having dangling files.
+class SelfDestructingPath {
+public:
+  explicit SelfDestructingPath(std::filesystem::path path);
+  ~SelfDestructingPath();
+  SelfDestructingPath(const SelfDestructingPath&) = delete;
+  SelfDestructingPath(SelfDestructingPath&&) = default;
+  auto operator=(const SelfDestructingPath&) -> SelfDestructingPath& = delete;
+  auto operator=(SelfDestructingPath&&) -> SelfDestructingPath& = default;
+
+  [[nodiscard]] static auto tempFile() -> SelfDestructingPath;
+
+  [[nodiscard]] static auto tempDir() -> SelfDestructingPath;
+
+  operator std::filesystem::path() const;  // NOLINT(google-explicit-conversion,
+                                           // google-explicit-constructor)
+
+  operator std::string() const;  // NOLINT(google-explicit-conversion,
+                                 // google-explicit-constructor)
+
+  operator std::string_view() const;  // NOLINT(google-explicit-conversion,
+                                      // google-explicit-constructor)
+
+private:
+  [[nodiscard]] static auto randomString() -> std::string;
+
+  std::filesystem::path path_;
+  std::string path_str_;
+};
+
+}  // namespace heph::utils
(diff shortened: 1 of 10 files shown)


modules/utils/include/hephaestus/utils/file.h
```diff
@@ -0,0 +1,44 @@
...
+namespace heph::utils {
+
+[[nodiscard]] auto readFile(const std::filesystem::path& path) -> std::string;
+[[nodiscard]] auto readBinaryFile(const std::filesystem::path& path) -> std::vector<std::byte>;
+
+void writeStringToFile(const std::filesystem::path& path, std::string_view content);
+void writeBufferToFile(const std::filesystem::path& path, std::span<std::byte> content);
+
+/// This class allows to create a file that is removed when the class goes out of scope (RAII).
+/// This is very useful in tests to avoid having dangling files.
```
lhruby, line 14: What speaks against `std::tmpfile`?

modules/utils/include/hephaestus/utils/file.h
```diff
@@ -0,0 +1,44 @@
...
+class SelfDestructingPath {
```
lhruby, line 15:
```suggestion
class TemporaryFile {
```

modules/utils/include/hephaestus/utils/file.h
```diff
@@ -0,0 +1,44 @@
...
+public:
+  explicit SelfDestructingPath(std::filesystem::path path);
+  ~SelfDestructingPath();
+  SelfDestructingPath(const SelfDestructingPath&) = delete;
+  SelfDestructingPath(SelfDestructingPath&&) = default;
+  auto operator=(const SelfDestructingPath&) -> SelfDestructingPath& = delete;
+  auto operator=(SelfDestructingPath&&) -> SelfDestructingPath& = default;
+
+  [[nodiscard]] static auto tempFile() -> SelfDestructingPath;
```
lhruby, line 24:
```suggestion
  [[nodiscard]] static auto createFile() -> SelfDestructingPath;
```

modules/utils/include/hephaestus/utils/file.h
```diff
@@ -0,0 +1,44 @@
...
+
+  [[nodiscard]] static auto tempDir() -> SelfDestructingPath;
```
lhruby, line 26:
```suggestion
  [[nodiscard]] static auto createPath() -> SelfDestructingPath;
```

modules/utils/src/file.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::error_code e;
+  std::filesystem::remove_all(path_, e);
+  if (e) {
+    fmt::print(stderr, "failed to remove test temporary directory {}, with error: {}\n", path_.string(),
+               e.message());
+  }
+}
+
+auto SelfDestructingPath::tempFile() -> SelfDestructingPath {
+  const auto global_temp_dir = std::filesystem::temp_directory_path();
```
lhruby, line 72:
```suggestion
  static const auto global_temp_dir = std::filesystem::temp_directory_path();
```

modules/utils/src/file.cpp
```diff
@@ -0,0 +1,101 @@
...
+  return SelfDestructingPath{ global_temp_dir / randomString() };
+}
+
+auto SelfDestructingPath::tempDir() -> SelfDestructingPath {
+  const auto global_temp_dir = std::filesystem::temp_directory_path();
```
lhruby, line 77:
```suggestion
  static const auto global_temp_dir = std::filesystem::temp_directory_path();
```

modules/utils/src/file.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::filesystem::remove_all(path_, e);
+  if (e) {
+    fmt::print(stderr, "failed to remove test temporary directory {}, with error: {}\n", path_.string(),
+               e.message());
+  }
+}
+
+auto SelfDestructingPath::tempFile() -> SelfDestructingPath {
+  const auto global_temp_dir = std::filesystem::temp_directory_path();
+  return SelfDestructingPath{ global_temp_dir / randomString() };
```
lhruby, line 73: Why not add a file extension, .tmp?

modules/utils/src/file.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+auto SelfDestructingPath::tempDir() -> SelfDestructingPath {
+  const auto global_temp_dir = std::filesystem::temp_directory_path();
+  auto temp_dir = global_temp_dir / randomString();
+  std::filesystem::create_directory(temp_dir);
```
lhruby, line 79: What if the folder exists already? Unlikely, but does it throw? Same for file actually.

modules/utils/include/hephaestus/utils/file.h
filippobrizzi: The reason why I didn't go for `TemporaryFile` is that in linux terms a temp file is created in `/tmp` which is emptied only on bootstrap. Here I wanted to make clear that the file is destroyed when you go out of scope.

But if you think `TemporaryFile` is more clear I will update it

modules/utils/include/hephaestus/utils/file.h
lhruby: sounds good

modules/utils/src/file.cpp
filippobrizzi: what's the point? it is not for humans to read

modules/utils/src/file.cpp
lhruby: My bad, I thought this was a const string.

modules/utils/src/file.cpp
lhruby: My bad, I thought this was a const string.

modules/utils/src/file.cpp
filippobrizzi: fixed it

modules/utils/src/file.cpp
lhruby: In case the file does not get deleted due to erroneous behavior, or if it is visible somewhere during build/testing.

It's nit, I just thought it would be a clean solution.

modules/utils/include/hephaestus/utils/file.h
lhruby: I like `TemporaryFile` tbh.

@floriantschopp Naming discussion, do you have an opinion?

floriantschopp: Agree with @filippobrizzi here, for a temporary file I would have assumed it to be in `/tmp` and handleded as Filippo described.

floriantschopp: How about volatile?

filippobrizzi: What about `ScopedPath`? like `std::scoped_lock`?

modules/utils/src/filesystem/file.cpp
```diff
@@ -0,0 +1,109 @@
...
+  // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+  infile.read(reinterpret_cast<char*>(buffer.data()), static_cast<std::streamsize>(file_size));
+
+  return buffer;
+}
+
+void writeStringToFile(const std::filesystem::path& path, std::string_view content) {
+  std::ofstream outfile{ path, std::ios::out };
+  throwExceptionIf<InvalidDataException>(!outfile,
+                                         fmt::format("Could not open file {} to write", path.string()));
```
lhruby, line 40: nit. Exception messages follow different conventions. Use read/write content both times or remove "content".

modules/utils/src/filesystem/file.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+auto ScopedPath::createDir() -> ScopedPath {
+  const auto global_temp_dir = std::filesystem::temp_directory_path();
+  auto temp_dir = global_temp_dir / randomString();
+  while (std::filesystem::exists(temp_dir)) {
+    temp_dir = global_temp_dir / randomString();
+  }
+
+  std::filesystem::create_directory(temp_dir);
+  return ScopedPath{ std::move(temp_dir) };
```
lhruby, line 88: Not sure on the role of path. Do you want to use path here, or is it a top level path? 
Furthermore, indicate if you want a global or relative path if it matters (probably it does not).

modules/utils/src/filesystem/file.cpp
```diff
@@ -0,0 +1,109 @@
...
+ScopedPath::ScopedPath(std::filesystem::path path) : path_(std::move(path)), path_str_(path_.string()) {
+}
+
+ScopedPath::~ScopedPath() {
+  std::error_code e;
+  std::filesystem::remove_all(path_, e);
+  if (e) {
+    fmt::print(stderr, "failed to remove test temporary directory {}, with error: {}\n", path_.string(),
+               e.message());
+  }
```
lhruby, line 67:
```suggestion
    fmt::print(stderr, "Failed to remove test temporary directory {}, with error: {}\n", path_.string(),
               e.message());
  }
```

modules/utils/tests/file_tests.cpp
```diff
@@ -0,0 +1,32 @@
...
+  const auto path = ScopedPath::createFile();
+  const auto content = std::vector<std::byte>{ std::byte{ 0x01 }, std::byte{ 0x02 }, std::byte{ 0x03 } };
+
+  writeBufferToFile(path, { content.data(), content.size() });
+  const auto read_content = readBinaryFile(path);
+
+  EXPECT_EQ(content, read_content);
+}
+
+TEST(File, ReadWrite) {
```
lhruby, line 23: nit. The first param of TEST should be a collection tag for tests, e.g. TEST(HephFilesystem, ...).

lhruby (approved):

modules/utils/src/filesystem/file.cpp
filippobrizzi: I am not sure I understand what you mean

modules/utils/include/hephaestus/utils/filesystem/file.h
```diff
@@ -0,0 +1,44 @@
...
+
+  operator std::string_view() const;  // NOLINT(google-explicit-conversion,
+                                      // google-explicit-constructor)
+
+private:
+  [[nodiscard]] static auto randomString() -> std::string;
+
+  std::filesystem::path path_;
+  std::string path_str_;
+};
```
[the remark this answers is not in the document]
filippobrizzi: done

PR #48, opened 2024-04-11
[Fix] includes

floriantschopp:
Some small fixes to includes, I am not sure why this even passed CI before tbh...

### DIFF
--- a/modules/utils/include/hephaestus/utils/filesystem/file.h
+++ b/modules/utils/include/hephaestus/utils/filesystem/file.h
@@ -1,6 +1,7 @@
 #pragma once
 #include <filesystem>
 #include <span>
+#include <vector>
 
 namespace heph::utils::filesystem {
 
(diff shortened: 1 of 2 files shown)


filippobrizzi (approved):

lhruby (approved):
Interesting. Are we using different compilers? gcc vs clang?

floriantschopp:
> Interesting. Are we using different compilers? gcc vs clang?

same compiler `clang++-18`, the previus shouldn't have compiled as it only included `<string>`, `std::transform` is not part of that...

lhruby:
I'm mainly wondering about `toupper` vs using the lambda you added. It compiled locally and in CI.

floriantschopp:
> I'm mainly wondering about `toupper` vs using the lambda you added. It compiled locally and in CI.

I got
```
stl_algo.h(4252, 5): Candidate template ignored: couldn't infer template argument '_UnaryOperation'
```

PR #52, opened 2024-04-16
[Feature] topic echo

floriantschopp:
# Description
* Move the example `sub_any` into an `ipc` executable
* Add option to truncate long items with `--noarr` like in `rostopic echo`

## Type of change


filippobrizzi (approved):
I tried this, and interestingly if you subscribe to a topic that doesn't exist you cannot terminate it with ctrl+c, I think the problem is that we get stuck waiting for something.
If you don't have time to look into it, can you add a TODO to look into it and I will try to check what is happening

floriantschopp:
replaced by https://github.com/olympus-robotics/hephaestus/pull/68

PR #61, opened 2024-05-08
Feat/Create random module.

lhruby:
# Description
The **random** module contains generic functionality to generate random data, as it is often used in unit tests.
* a random number generator (RNG) which is pre-configured to be deterministic or not in order to enforce (non-)determinism in all tests
* a method `randomT<T>(std::mt19937_64& mt) -> T` which can be used to create a random element of a primitive, custom or container type

## Type of change


modules/testing/include/hephaestus/testing/random_container.h
```diff
@@ -0,0 +1,45 @@
...
+#include "hephaestus/testing/random_type.h"
+
+namespace heph::testing {
+
+//=================================================================================================
+// Random vector generation
+//=================================================================================================
+namespace internal {
+template <typename T>
+concept IsVectorT =
```
filippobrizzi, line 21: nit. this feels like a quite generic concept, maybe we can move it to a general location

filippobrizzi (approved):
This is fantastic! I would however rename the module to `random` as it could be useful not only for testing.

modules/testing/include/hephaestus/testing/random_container.h
lhruby: Talked this out in person. We'll leave generalizing this concept to the time when there is a proper traits module.

PR #65, opened 2024-05-22
Zenoh bag recorder: refactor to use DynamicSubscriber

filippobrizzi:
# Description
Refactor zenoh bag recorder to use DynamicSubscriber.


floriantschopp (approved):
lgtm

PR #68, opened 2024-05-24
[Feature] noarr for topic echo 

floriantschopp:
# Description
Replaces https://github.com/olympus-robotics/hephaestus/pull/52

* Add option to truncate long items with --noarr like in rostopic echo
* Move from examples to executable
## Type of change

### DIFF
--- a/modules/ipc/apps/CMakeLists.txt
+++ b/modules/ipc/apps/CMakeLists.txt
@@ -23,7 +23,7 @@ define_module_executable(
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
         PUBLIC_LINK_LIBS "")
 
-define_module_example(
+define_module_executable(
         NAME zenoh_topic_echo
         SOURCES zenoh_topic_echo.cpp zenoh_program_options.h
         PUBLIC_INCLUDE_PATHS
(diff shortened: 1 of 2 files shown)


modules/ipc/apps/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ define_module_executable(
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
         PUBLIC_LINK_LIBS "")
 
-define_module_example(
+define_module_executable(
```
filippobrizzi, line 26: great catch!

modules/ipc/apps/zenoh_topic_echo.cpp
```diff
@@ -21,9 +22,35 @@
 #include "zenoh_program_options.h"
 
 namespace heph::ipc::apps {
+
+namespace {
+constexpr auto MAX_ARRAY_LENGTH = 100;
```
filippobrizzi, line 27: maybe this could be a paramter? 100 seems like a lot.

we could have a flag: `--arr-max-len` with default value 0 or -1 for infinite

filippobrizzi (approved):

modules/ipc/apps/zenoh_topic_echo.cpp
floriantschopp: how about this? the benefit is that it supports similar usage as with ROS...

PR #72, opened 2024-06-03
Zenoh bag player

filippobrizzi:
# Description
Add bag player library and app:
*  Right now the bag player is basic and all topic share the same thread. This I already know is a problem but I wanted to have something running end-to-end
* Add test that: create an mcap bag -> play the bag -> record the bag -> check that the two bags are the same
  * To do this I create test protofile in the bag test folder. These are very similar to the one that already exists in the `serdes` module. I will unify them later as I need to find a satisfying way of doing it.
* Extendend signal handler utilis to support   the case of bag playback where we want to terminate either because the app complete or via signal, please provide better naming!

### DIFF
--- a/modules/bag/apps/bag_player.cpp
+++ b/modules/bag/apps/bag_player.cpp
@@ -1,0 +1,59 @@
+//=================================================================================================
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#include <csignal>
+#include <exception>
+#include <filesystem>
+#include <future>
+
+#include <absl/log/log.h>
+#include <fmt/core.h>
+#include <mcap/reader.hpp>
+
+#include "hephaestus/bag/zenoh_player.h"
+#include "hephaestus/cli/program_options.h"
+#include "hephaestus/ipc/program_options.h"
+#include "hephaestus/utils/exception.h"
+#include "hephaestus/utils/signal_handler.h"
+#include "hephaestus/utils/stack_trace.h"
+
+auto main(int argc, const char* argv[]) -> int {
+  heph::utils::StackTrace stack_trace;
+
+  try {
+    auto desc = heph::cli::ProgramDescription("Playback a bag to zenoh topics");
+    heph::ipc::appendIPCProgramOption(desc);
+    desc.defineOption<std::filesystem::path>("input_bag", 'i', "output file where to write the bag")
+        .defineFlag("wait_for_readers_to_connect", 'w',
+                    "Wait for readers to connect before starting playback");
+    const auto args = std::move(desc).parse(argc, argv);
+    auto input_file = args.getOption<std::filesystem::path>("input_bag");
+    auto wait_for_readers_to_connect = args.getOption<bool>("wait_for_readers_to_connect");
+    auto [config, _] = heph::ipc::parseIPCProgramOptions(args);
+
+    LOG(INFO) << fmt::format("Reading bag file: {}", input_file.string());
+
+    heph::throwExceptionIf<heph::InvalidDataException>(
+        !std::filesystem::exists(input_file),
+        fmt::format("input bag file {} doesn't exist", input_file.string()));
+    auto bag_reader = std::make_unique<mcap::McapReader>();
+    const auto status = bag_reader->open(input_file.string());
+    if (status.code != mcap::StatusCode::Success) {
+      fmt::print(stderr, "Failed to open bag file: {}", status.message);
+      std::exit(1);
+    }
+
+    heph::bag::ZenohPlayerParams params{ .session = heph::ipc::zenoh::createSession(std::move(config)),
+                                         .bag_reader = std::move(bag_reader),
+                                         .wait_for_readers_to_connect = wait_for_readers_to_connect };
+    auto zenoh_player = heph::bag::ZenohPlayer::create(std::move(params));
+    zenoh_player.start().get();
+
+    heph::utils::SignalHandlerStopOrAppComplete::wait(zenoh_player);
+
+  } catch (std::exception& e) {
+    fmt::println("Failed with exception: {}", e.what());
+    std::exit(1);
+  }
+}
--- a/modules/bag/apps/bag_player.cpp
+++ b/modules/bag/apps/bag_player.cpp
@@ -50,7 +50,7 @@ auto main(int argc, const char* argv[]) -> int {
     auto zenoh_player = heph::bag::ZenohPlayer::create(std::move(params));
     zenoh_player.start().get();
 
-    heph::utils::SignalHandlerStopOrAppComplete::wait(zenoh_player);
+    heph::utils::InterruptHandlerOrAppComplete::wait(zenoh_player);
 
   } catch (std::exception& e) {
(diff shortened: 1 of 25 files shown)


modules/utils/include/hephaestus/utils/signal_handler.h
```diff
@@ -33,4 +34,58 @@ class SignalHandlerStop {
...
+/// Usage:
+/// ```
+/// int main() {
+///   auto app = MyApp{};
+///   app.start().get();
+///   SignalHandlerStopOrAppComplete::wait(app);
+/// }
+/// ```
+/// > NOTE: `MayApp` needs to satify the stoppable compoenent interface.
+class SignalHandlerStopOrAppComplete {
```
filippobrizzi, line 50: Please help me find a better name!!

modules/bag/src/zenoh_player.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    const auto& topic = message.channel->topic;
+    const auto& publisher = publishers_[topic];
+
+    const auto current_msg_timestamp =
+        std::chrono::steady_clock::time_point{ std::chrono::nanoseconds{ message.message.publishTime } };
+
+    const auto write_timestamp = current_msg_timestamp - first_msg_timestamp + first_playback_timestamp;
+
+    if (auto now = std::chrono::steady_clock::now(); now > write_timestamp && msgs_played_count > 0) {
```
floriantschopp, line 147: const?

modules/bag/src/zenoh_player.cpp
```diff
@@ -0,0 +1,191 @@
...
+      ++deadline_missed_count;
+      LOG(WARNING) << fmt::format("failed to publish message {} in time for topic {}, delay {}",
```
floriantschopp, line 149: I would change this error message, to `deadline missed on message...`

modules/bag/tests/test_proto_conversion.h
```diff
@@ -0,0 +1,73 @@
...
+  auto operator==(const Company& other) const -> bool = default;
+  [[nodiscard]] static auto random(std::mt19937_64& mt) -> Company;
+
+  std::string name;
+  int employer_count{};
+};
+
+auto Company::random(std::mt19937_64& mt) -> Company {
+  return { .name = random::randomT<std::string>(mt), .employer_count = random::randomT<int>(mt) };
+}
```
floriantschopp, line 35: nit: kinda weird examples for the test in  a robotics scenario :)

modules/bag/tests/test_proto_conversion.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
```
floriantschopp, line 1: i don't see any tests here, should this move to `examples`?

modules/utils/include/hephaestus/utils/signal_handler.h
```diff
@@ -39,4 +42,59 @@ class InterruptHandler {
...
+/// \brief Use this class to block until a signal is received or the application completes.
+/// Usage:
+/// ```
+/// int main() {
+///   auto app = MyApp{};
+///   app.start().get();
+///   InterruptHandlerOrAppComplete::wait(app);
+/// }
+/// ```
+/// > NOTE: `MayApp` needs to satify the stoppable compoenent interface.
```
floriantschopp, line 57:
```suggestion
/// > NOTE: `MyApp` needs to satify the stoppable compoenent interface.
```

modules/utils/include/hephaestus/utils/signal_handler.h
floriantschopp: why not `StopableAndWaitableApp`?

modules/bag/apps/bag_player.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+auto main(int argc, const char* argv[]) -> int {
+  heph::utils::StackTrace stack_trace;
+
+  try {
+    auto desc = heph::cli::ProgramDescription("Playback a bag to zenoh topics");
+    heph::ipc::appendIPCProgramOption(desc);
+    desc.defineOption<std::filesystem::path>("input_bag", 'i', "output file where to write the bag")
+        .defineFlag("wait_for_readers_to_connect", 'w',
+                    "Wait for readers to connect before starting playback");
```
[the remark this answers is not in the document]
filippobrizzi: this is one wait for a reader on `every` topic. This feature is very useful for testing (to make it deterministic) and also to make running experiments more deterministic. We were using similar concept a lot in arrival as well

modules/bag/include/hephaestus/bag/zenoh_player.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include <future>
+
+#include "hephaestus/ipc/zenoh/session.h"
+
+#define MCAP_IMPLEMENTATION
+#define MCAP_COMPRESSION_NO_ZSTD
+#define MCAP_COMPRESSION_NO_LZ4
```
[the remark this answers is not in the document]
filippobrizzi: I think here I would apply a Go proverb:
[A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

modules/bag/src/zenoh_player.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  LOG(INFO) << fmt::format("Created publisher for topic: {}", channel.topic);
+}
+
+void ZenohPlayer::Impl::run() {
+  auto read_options = mcap::ReadMessageOptions{};
+  read_options.readOrder = mcap::ReadMessageOptions::ReadOrder::LogTimeOrder;
+  auto messages = bag_reader_->readMessages([](const auto&) {}, read_options);
+
+  auto first_msg_timestamp = std::chrono::steady_clock::time_point{ std::chrono::nanoseconds{
```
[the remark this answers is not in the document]
filippobrizzi: you are right, changed to system clock

modules/bag/src/zenoh_player.cpp
filippobrizzi: nice!

modules/bag/tests/test_proto_conversion.h
filippobrizzi: good point, renamed

modules/bag/tests/test_proto_conversion.h
filippobrizzi: This file is used only for testing. If I move it to examples it won't work.

I would probably need to create a `test_utils` module where I put this so that it can be reused, but the problem is that current CMake structure doesn't allow to create a dependency only for testing so the module would be installed everytime even when building heph as a dependency.

Now this is probably not a problem, but it is not the cleanest solution.

modules/utils/include/hephaestus/utils/signal_handler.h
filippobrizzi: I asked ChatGPT and it says: `SignalAwaiter` which made me think that maybe I can just merge the two classes and provide:
```
static void wait();
template <StoppableAndWaitable T>
static void wait(T& app);
```

PR #73, opened 2024-06-03
Stacktrace print util

filippobrizzi:
# Description
Create a class util that pretty prints the stack trace when there is a segfault.

Add the class to all app binary.


johannes-graeter:
sorry for hooking in unasked, but would https://en.cppreference.com/w/cpp/utility/basic_stacktrace an alternative (voting for c++23 :) )

filippobrizzi:
> sorry for hooking in unasked, but would https://en.cppreference.com/w/cpp/utility/basic_stacktrace an alternative (voting for c++23 :) )

Yep, as soon as we can switch to C++23 we can replace the print with that. The class is still useful to abstract the signal registration

floriantschopp (approved):
great stuff :muscle:

PR #75, opened 2024-06-03
Fix: various minor zenoh fixes

filippobrizzi:
# Description
* `PublisherDiscovery` was not correctly shutting down zenoh data
* `Publisher` move constructor doesn't work (probably due to some internal zenoh limitation) so we are deleting it
* `publish` now accept const data to be more permissive


lhruby (approved):
LGTM. Good solution to remove the publisher move constructor.

floriantschopp (approved):

PR #82, opened 2024-06-19
Watchdog timer

filippobrizzi:
# Description
Create a watchdog timer. 

The timer call the given callback every `period` time.
A function `pat()` is also provided to implement a dead-man switch.

Example use case:
In a motor controller, if no command is received within the time period, stop the motors.

### DIFF
--- a/modules/utils/src/timing/watchdog.cpp
+++ b/modules/utils/src/timing/watchdog.cpp
@@ -1,0 +1,45 @@
+//=================================================================================================
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#include "hephaestus/utils/timing/watchdog.h"
+
+namespace heph::utils::timing {
+void WatchdogTimer::start(std::chrono::milliseconds period, Callback&& callback) {
+  period_ = period;
+  callback_ = std::move(callback);
+  running_ = true;
+  pat_ = false;
+  start_timestamp_ = std::chrono::system_clock::now();
+  thread_ = std::thread([this]() { loop(); });
+}
+
+void WatchdogTimer::stop() {
+  running_ = false;
+  condition_.notify_all();
+  if (thread_.joinable()) {
+    thread_.join();
+  }
+}
+
+void WatchdogTimer::pat() {
+  pat_ = true;
+}
+
+void WatchdogTimer::loop() {
+  while (running_) {
+    std::unique_lock<std::mutex> lock(mutex_);
+    const auto target_timestamp = start_timestamp_ + (++fire_count_) * period_;
+    if (condition_.wait_until(lock, target_timestamp, [this] { return !running_; })) {
+      break;
+    }
+
+    if (pat_) {
+      pat_ = false;
+      continue;
+    }
+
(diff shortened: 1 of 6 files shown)


modules/utils/src/timing/watchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+  pat_ = true;
+}
+
+void WatchdogTimer::loop() {
+  while (running_) {
+    std::unique_lock<std::mutex> lock(mutex_);
+    const auto target_timestamp = start_timestamp_ + (++fire_count_) * period_;
+    if (condition_.wait_until(lock, target_timestamp, [this] { return !running_; })) {
+      break;
+    }
```
lhruby, line 35: Beautiful!

modules/utils/tests/watchdog_tests.cpp
```diff
@@ -0,0 +1,42 @@
...
+TEST(Watchdog, TimerFiring) {
+  constexpr auto PERIOD = std::chrono::milliseconds{ 10 };
+  WatchdogTimer timer;
+  std::atomic<int> count{ 0 };
+  timer.start(PERIOD, [&count]() { ++count; });
+  EXPECT_EQ(count, 0);
+  std::this_thread::sleep_for(PERIOD * 4);
+  timer.stop();
+
+  EXPECT_GE(count, 3);
```
lhruby, line 22: Are you sure that stop and wakeup calls will not create jitter in the execution time of the two threads? If so, you might want to increase the PERIOD from 10 ms to e.g. 100 ms. Only increase PERIOD if really necessary, since this will increase the overall unit test time.

lhruby (approved):
Great work.

modules/utils/src/timing/watchdog.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "hephaestus/utils/timing/watchdog.h"
+
+namespace heph::utils::timing {
+void WatchdogTimer::start(std::chrono::milliseconds period, Callback&& callback) {
+  period_ = period;
+  callback_ = std::move(callback);
+  running_ = true;
+  pat_ = false;
+  start_timestamp_ = std::chrono::system_clock::now();
```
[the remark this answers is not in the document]
filippobrizzi: not really cause if you notice I use `wait_until` so I want to be aligned with the real clock, and not have drifts

modules/utils/tests/watchdog_tests.cpp
filippobrizzi: Good point, this is why I am using `  EXPECT_GE` instead of exact values, as it is not deterministic.

modules/utils/src/timing/watchdog.cpp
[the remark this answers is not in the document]
lhruby: Just read the docs. I have to admit I don't understand it 100%, but it seems like they take care of jumps internally as much as possible, and that std::steady_clock would still suffer from system time jumps. Makes sense, as they compare to absolute time. 

So, keep system_clock, it's good as is.

PR #83, opened 2024-06-19
Add BitFlag class

filippobrizzi:
# Description
Add a new class `BitFlag<EnumT>` that allows to create bit flags.

The usage is visible from the unit tests and the class documentation. If something is not clear, please let me know I will improve the feedback.


modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,96 @@
...
+
+  /// Unset the given flag
+  constexpr auto unset(EnumT flag) -> BitFlag& {
+    value_ =
+        static_cast<EnumT>(static_cast<T>(value_) & ~static_cast<T>(flag));  // NOLINT(hicpp-signed-bitwise)
+    return *this;
+  }
+
+  /// Retunrs true if the input flag is set
+  [[nodiscard]] constexpr auto has(EnumT flag) const -> bool {
```
lhruby, line 73: Does a bit pattern "have" flags, or are flags "set"?

I would naiively expect isSet instead of has.

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,96 @@
...
+  // see https://dietertack.medium.com/using-bit-flags-in-c-d39ec6e30f08
+  using T = std::underlying_type<EnumT>::type;
+
+  constexpr explicit BitFlag(EnumT value) : value_{ value } {
+    static_assert(checkEnumValuesArePowerOf2<EnumT>(),
+                  "Enum is not valid for BitFlag, its values must be power of 2.");
+  }
+
+  /// Unset all flags and set the given flag.
+  constexpr auto reset(EnumT flag) -> BitFlag& {
```
lhruby, line 54: What is the use case of reset all, and then set one, in one function call? Alternative would be reset(), and then set(flag)

modules/utils/tests/bit_flag_tests.cpp
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_TRUE(flag.has(TestEnum::D));
+}
+
+TEST(BitFlag, SetMultiple) {
+  BitFlag<TestEnum> flag{ TestEnum::A };
+  static constexpr auto E = BitFlag<TestEnum>{ TestEnum::C }.set(TestEnum::D);
+  flag.set(E);
+  EXPECT_TRUE(flag.has(TestEnum::C));
+  EXPECT_TRUE(flag.has(TestEnum::D));
+}
```
lhruby, line 53: Redundant. Retest of loc 41-44.

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,96 @@
...
+
+  /// Retunrs true if the input flag is the only one set
+  [[nodiscard]] constexpr auto hasOnly(EnumT flag) const -> bool {
+    return (static_cast<T>(value_) & static_cast<T>(flag)) == static_cast<T>(value_);
+  }
+
+  /// Retunrs true if any of the input flags are set
+  [[nodiscard]] constexpr auto hasAny(EnumT flag) const -> bool {
+    return (static_cast<T>(value_) & static_cast<T>(flag)) != 0u;
+  }
```
lhruby, line 85: Given you implemented hasAny, you might also want to add `hasAll`

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,96 @@
...
+    return (static_cast<T>(value_) & static_cast<T>(flag)) == static_cast<T>(flag);
+  }
+
+  /// Retunrs true if the input flag is the only one set
+  [[nodiscard]] constexpr auto hasOnly(EnumT flag) const -> bool {
+    return (static_cast<T>(value_) & static_cast<T>(flag)) == static_cast<T>(value_);
+  }
+
+  /// Retunrs true if any of the input flags are set
+  [[nodiscard]] constexpr auto hasAny(EnumT flag) const -> bool {
```
lhruby, line 83: nit. `hasAny` --> `hasAnyOf`. The name `hasAny` suggests a functionality like `!isEmpty`

modules/utils/include/hephaestus/utils/bit_flag.h
lhruby: Resolved in conversation. `has` seems to be the commonly used term.

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,101 @@
...
+    return *this;
+  }
+
+  /// Unset the given flag(s)
+  constexpr auto unset(const BitFlag& flag) -> BitFlag& {
+    value_ = value_ & static_cast<T>(~flag.value_);
+    return *this;
+  }
+
+  /// Retunrs true if the input flag(s) is set
```
lhruby, line 81:
```suggestion
  /// Returns true if the input flag(s) is set
```

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,101 @@
...
+  [[nodiscard]] constexpr auto has(const BitFlag& flag) const -> bool {
+    return (value_ & flag.value_) == flag.value_;
+  }
+
+  /// Retunrs true if the input flag is the only one set
+  [[nodiscard]] constexpr auto hasExactly(const BitFlag& flag) const -> bool {
+    return (value_ & flag.value_) == value_;
+  }
+
+  /// Retunrs true if any of the input flags are set
```
lhruby, line 91:
```suggestion
  /// Returns true if the input flag is the only one set
  [[nodiscard]] constexpr auto hasExactly(const BitFlag& flag) const -> bool {
    return (value_ & flag.value_) == value_;
  }

  /// Returns true if any of the input flags are set
```

modules/utils/include/hephaestus/utils/bit_flag.h
filippobrizzi: I thought about that, but the flag needs to have at least one value, you cannot be empty, can you?

modules/utils/include/hephaestus/utils/bit_flag.h
filippobrizzi: `hasAll` is the same as `has`

modules/utils/include/hephaestus/utils/bit_flag.h
filippobrizzi: well but we pass a flag, so I think it is clear that it is related to the flag, but I can update then name. Usually you know I am all in favor of long descriptive names, but here I am actually trying to keep consistency for the different functions

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  /// Unset all flags and set the given flag.
+  constexpr auto reset(EnumT flag) -> BitFlag& {
+    value_ = flag;
+    return *this;
+  }
+
+  /// Set the input flag
+  constexpr auto set(EnumT flag) -> BitFlag& {
```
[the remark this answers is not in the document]
filippobrizzi: you are right

modules/utils/include/hephaestus/utils/bit_flag.h
lhruby: Unless you have different information, I don't think any flag has to be set. If each flag signals an error, no flag means operational as desired. In that light, I would reset to 0. The current reset can nicely be achieved via reset().set(A) then

modules/utils/include/hephaestus/utils/bit_flag.h
lhruby: My reasoning is mainly inspired by https://en.cppreference.com/w/cpp/algorithm/all_any_none_of. Both names are fine, foo_of just is closer to similar stl algorithms.

modules/utils/include/hephaestus/utils/bit_flag.h
filippobrizzi: actually, 0 is a valid value, that is correct, also if I set A and then unset A I would have no value. Updating

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,103 @@
...
+    return *this;
+  }
+
+  /// Unset the given flag(s)
+  constexpr auto unset(BitFlag flag) -> BitFlag& {
+    value_ = value_ & static_cast<T>(~flag.value_);
+    return *this;
+  }
+
+  /// Returns true if the input flag(s) is set
```
lhruby, line 83:
```suggestion
  /// Returns true if the input flag is set. In case of multiple input flags, returns true if all input flags are set.
```

modules/utils/include/hephaestus/utils/bit_flag.h
```diff
@@ -0,0 +1,103 @@
...
+  [[nodiscard]] constexpr auto has(BitFlag flag) const -> bool {
+    return (value_ & flag.value_) == flag.value_;
+  }
+
+  /// Returns true if the input flag is the only one set
+  [[nodiscard]] constexpr auto hasExactly(BitFlag flag) const -> bool {
+    return (value_ & flag.value_) == value_;
+  }
+
+  /// Returns true if any of the input flags are set
```
lhruby, line 93:
```suggestion
  /// Returns true if any of the input flags are set. In case of a single input flag, hasAny is equivalent to has.
```

modules/utils/include/hephaestus/utils/bit_flag.h
lhruby: true. I added a suggestion to extend the comment. No further requests.

lhruby (approved):

PR #85, opened 2024-06-21
CMake: update MCAP installation

filippobrizzi:
# Description
After, discussing with the MCAP author they decided not to proceed with my proposal of adding CMake support to mcap (https://github.com/foxglove/mcap/pull/1065/files). 

For this reason, I decided to go down a different route and add the cmake support to a separate repo that will work as a wrapper (https://github.com/olympus-robotics/mcap_builder). 

This also simplifies MCAP usability.


floriantschopp (approved):
nice, lgtm

PR #98, opened 2024-06-25
Fix some IWYU issues; fix typo in README.

hzeller:
# Description

  * Fix some missing includes to make things compile; it looks like there might be other instances where more diligence is needed w.r.t IWYU, but these I stumbled upon as it made the compile hick-up.
  * The `ninja checks` in the README should really be `check`
  * There is another typo in the README `cmake --preset preferred` ... the `preferred` preset does not actually exist. Maybe `default` was meant ? Anyway, not touching that, but you should probably change that to one of the existing options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ ninja all examples
 
 To compile and run the unit test
 ```bash
-ninja checks
+ninja check
 ```
 
 > TODO: add section on the different flags and options.
--- a/modules/utils/include/hephaestus/utils/timing/watchdog.h
+++ b/modules/utils/include/hephaestus/utils/timing/watchdog.h
@@ -5,6 +5,8 @@
 #pragma once
 
 #include <chrono>
+#include <condition_variable>
+#include <cstdint>
 #include <functional>
 #include <thread>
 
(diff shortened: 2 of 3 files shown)


filippobrizzi:
@hzeller Thanks a lot for your contribution!
You are right, `preferred` is now `default`, I will update the readme.

After your comment, I found that there is this tool [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use) (which you probably know) and I integrated it in the repo https://github.com/olympus-robotics/hephaestus/pull/99, although I still need to understand the proper configuration for it.

filippobrizzi (approved):

hzeller:
Since you're using `clang-tidy` in the project, enabling the `misc-include-cleaner` check might already provide a bunch of detection without having to use include-what-you-use.

Also the clang layering check is helpful, there are a few options that help keeping include dependencies clean: https://maskray.me/blog/2022-09-25-layering-check-with-clang

filippobrizzi:
Thanks for the tip `misc-include-cleaner` provided to be more helpful than `include-what-you-use`

I have enabled it for the repo and found a lot of missing includes https://github.com/olympus-robotics/hephaestus/pull/107

hzeller:
Cool!

PR #101, opened 2024-06-25
[Tests] add tests for IPC

floriantschopp:
# Description

Adds a unit test for pub/sub and service/client usage.

## Type of change

### DIFF
--- a/modules/examples/tests/zenoh_tests.cpp
+++ b/modules/examples/tests/zenoh_tests.cpp
@@ -44,21 +44,23 @@ TEST(ZenohTests, MessageExchange) {
   EXPECT_EQ(send_message, received_message);
 }
 
-TEST(IPCExampleClientTest, ServiceCallExchange) {
+TEST(ZenohTests, ServiceCallExchange) {
   std::mt19937_64 mt{ kSeed };  // NOLINT
 
   const auto request_message = randomPose(mt);
   auto response_message = randomPose(mt);
-  EXPECT_NE(response_message, response_message);
+  EXPECT_NE(request_message, response_message);
 
   const auto service_topic = ipc::TopicConfig("test_service");
-  ipc::Config config{};
-  auto session = ipc::zenoh::createSession(std::move(config));
 
-  auto service_server =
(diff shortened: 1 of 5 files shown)


modules/examples/tests/zenoh_tests.cpp
```diff
@@ -0,0 +1,72 @@
...
+  auto subscriber = ipc::subscribe<heph::ipc::zenoh::Subscriber, Pose>(
+      session, topic,
+      [&received_message]([[maybe_unused]] const ipc::MessageMetadata& metadata,
+                          const std::shared_ptr<Pose>& message) { received_message = *message; });
+
+  const auto success = publisher.publish(send_message);
+  EXPECT_TRUE(success);
+
+  // Give enough time for the message to be received
+  std::this_thread::sleep_for(std::chrono::seconds(1));
```
filippobrizzi, line 43: I think 1ms should be enough here

modules/examples/tests/zenoh_tests.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  ipc::Config server_config{};
+  auto server_session = ipc::zenoh::createSession(std::move(server_config));
+  auto service_server = ipc::zenoh::Service<Pose, Pose>(server_session, service_topic,
+                                                        [](const Pose& request) { return request; });
+
+  ipc::Config client_config{};
+  auto client_session = ipc::zenoh::createSession(std::move(client_config));
+  const auto reply = ipc::zenoh::callService<Pose, Pose>(*client_session, service_topic, request_message,
+                                                         std::chrono::milliseconds(1000));
```
filippobrizzi, line 65: also here the timeout it is too long

filippobrizzi (commented):
this test should go in the `ipc` package, but to do that we would need to first create the `test_helper` package to provide with dummy proto messages.
I would say, let's leave it here with a comment and let's add a task in the project to move it when we have the test helper

modules/examples/tests/zenoh_tests.cpp
floriantschopp: 1 ms was not long enough...

filippobrizzi (approved):

PR #103, opened 2024-06-26
Spinner add option to have a loop period

filippobrizzi:
# Description
Add option to have a loop period that allow to execute the callback at a given rate.


floriantschopp:
Good stuff! Nit: could we move to frequency instead of period?

floriantschopp (approved):

floriantschopp (approved):

PR #76, opened 2024-06-05
Fix: I hate CMake!!  

filippobrizzi:
# Description
Fix how proto generated files and library are installed.

PR #104, opened 2024-06-26
[Fix] Make compatible with cmake 3.22

floriantschopp:
# Description
https://github.com/olympus-robotics/hephaestus/pull/76 made a dependency on cmake 3.29 which has some core changes and might not be supported on all systems. This PR makes the build compatible with 3.22 in general, except for the proto install (warning issued).

## Type of change


cmake/external.cmake
```diff
@@ -90,6 +90,11 @@ macro(add_cmake_dependency)
       add_dummy_target(${TARGET_ARG_NAME})
     else()
       message(STATUS "    ${TARGET_ARG_NAME}: Building ${TARGET_ARG_VERSION} from source")
+      if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.29")
+          set(download_extract_timestamp_arg "DOWNLOAD_EXTRACT_TIMESTAMP;TRUE")
```
filippobrizzi, line 94: nit. 
```suggestion
          set(DOWNLOAD_EXTRACT_TIMESTAMP_ARG "DOWNLOAD_EXTRACT_TIMESTAMP;TRUE")
```

filippobrizzi (approved):

PR #107, opened 2024-06-28
Clang tidy enable include check

filippobrizzi:
# Description
Enabled `misc-*` clang-tidy checks (except `misc-no-recursion` ).

This mainly added two checks: include what you use and forcing variables to be const when possible.


floriantschopp (approved):
lgtm

PR #108, opened 2024-07-01
[cmake] disable exception flag

floriantschopp:
To be able to comply with google style, no exceptions should be thrown in the code (https://google.github.io/styleguide/cppguide.html#Exceptions). Therefore, I added an option to disable exceptions and replace with the `CHECK` macro of abseil.


johannes-graeter:
Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.

cmake/03_compiler_config.cmake
```diff
@@ -171,6 +171,12 @@ if(ENABLE_LINTER)
   endif()
 endif()
 
+# Configure test coverage
+option(DISABLE_EXCEPTIONS "Disable exceptions and use CHECK instead" OFF)
+if(DISABLE_EXCEPTIONS)
+  add_definitions(-DDISABLE_EXCEPTIONS)
```
filippobrizzi, line 177: `add_definitions` has been deprecateed
https://cmake.org/cmake/help/latest/command/add_definitions.html
```suggestion
  add_compile_definitions(-DDISABLE_EXCEPTIONS)
```

modules/utils/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ set(SOURCES
 # library target
 define_module_library(
   NAME utils
-  PUBLIC_LINK_LIBS absl::random_random absl::time fmt::fmt
+  PUBLIC_LINK_LIBS absl::random_random absl::time fmt::fmt absl::log
```
filippobrizzi, line 50: Could you please add also, I forgot it from the bit flag PR
```suggestion
  PUBLIC_LINK_LIBS absl::random_random absl::time fmt::fmt absl::log magic_enum::magic_enum
```

filippobrizzi (approved):

filippobrizzi:
> Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.

But then when you are actually throwing exception it is not clear

johannes-graeter:
> > Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.
> 
> But then when you are actually throwing exception it is not clear

True, this would IMO only make sense, when not catching the exceptions in most cases (like go that has "recover" in cases you really want to recover from a panic...).
On the other hand, it is strange when throwException does not emit an exception (caused by a compiler flag, which might not be visible from the code). So I guess a different name would be appropriate.
In absence of a better idea I would suggest maybeThrowException :)
Since I am missing some of the context, it might also too much nitpicking from my side ;)

filippobrizzi:
> > > Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.
> > 
> > 
> > But then when you are actually throwing exception it is not clear
> 
> True, this would IMO only make sense, when not catching the exceptions in most cases (like go that has "recover" in cases you really want to recover from a panic...). On the other hand, it is strange when throwException does not emit an exception (caused by a compiler flag, which might not be visible from the code). So I guess a different name would be appropriate. In absence of a better idea I would suggest maybeThrowException :) Since I am missing some of the context, it might also too much nitpicking from my side ;)

@johannes-graeter Given that the default behaviour is to throw exceptions and to switch to the other behaviour you need to set that you want exceptions disabled explicitly I would keep this name. If you have manually disabled exception you know that you are not really throwing.

johannes-graeter:
> > > > Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.
> > > 
> > > 
> > > But then when you are actually throwing exception it is not clear
> > 
> > 
> > True, this would IMO only make sense, when not catching the exceptions in most cases (like go that has "recover" in cases you really want to recover from a panic...). On the other hand, it is strange when throwException does not emit an exception (caused by a compiler flag, which might not be visible from the code). So I guess a different name would be appropriate. In absence of a better idea I would suggest maybeThrowException :) Since I am missing some of the context, it might also too much nitpicking from my side ;)
> 
> @johannes-graeter Given that the default behaviour is to throw exceptions and to switch to the other behaviour you need to set that you want exceptions disabled explicitly I would keep this name. If you have manually disabled exception you know that you are not really throwing.

@filippobrizzi I see your point and agree that this is good for now since users compile hephaestus themselfs. I still think this might open some problems in the future, since the contract for error handling depends on compiler flags instead of code. If released in a compiled state, it should be certain, that this gets shipped only with one version of the state. Users could otherwise have a hard time finding out why their recovery (from a wrong filename with try/catch f.e. ) results in a fatal...

johannes-graeter:
Thanks for your patience @floriantschopp and @filippobrizzi, error handling is a very dear topic to me so I am very picky about it(since I have wasted soooo many hours coping with error handling strategies of external libs...)

filippobrizzi:
> > > > > Great! so the exceptions will effectively be turned into what other languages call a panic right? Perhaps it would then make sense to rename "throwException" to "panic" to empahsize this.
> > > > 
> > > > 
> > > > But then when you are actually throwing exception it is not clear
> > > 
> > > 
> > > True, this would IMO only make sense, when not catching the exceptions in most cases (like go that has "recover" in cases you really want to recover from a panic...). On the other hand, it is strange when throwException does not emit an exception (caused by a compiler flag, which might not be visible from the code). So I guess a different name would be appropriate. In absence of a better idea I would suggest maybeThrowException :) Since I am missing some of the context, it might also too much nitpicking from my side ;)
> > 
> > 
> > @johannes-graeter Given that the default behaviour is to throw exceptions and to switch to the other behaviour you need to set that you want exceptions disabled explicitly I would keep this name. If you have manually disabled exception you know that you are not really throwing.
> 
> @filippobrizzi I see your point and agree that this is good for now since users compile hephaestus themselfs. I still think this might open some problems in the future, since the contract for error handling depends on compiler flags instead of code. If released in a compiled state, it should be certain, that this gets shipped only with one version of the state. Users could otherwise have a hard time finding out why their recovery (from a wrong filename with try/catch f.e. ) results in a fatal...

cmake/03_compiler_config.cmake
```diff
@@ -171,6 +171,12 @@ if(ENABLE_LINTER)
   endif()
 endif()
 
+# This will deactivate the use of exceptions and use CHECK instead. The whole code should be considered noexcept.
```
filippobrizzi, line 174:
```suggestion
# This will deactivate the use of exception, and the utility functions we use to wrap throwing exceptions will instead terminate displaying an error message. The whole code should be considered `noexcept`.
```

modules/utils/include/hephaestus/utils/exception.h
```diff
@@ -21,24 +24,43 @@ class Exception : public std::runtime_error {
   explicit Exception(const std::string& message, std::source_location location);
 };
 
-/// @brief  User function to throw an exception
+/// @brief  User function to throw an exception. The whole code should be considered noexcept. Will use CHECK
+/// if DISABLE_EXCEPTIONS = ON
```
filippobrizzi, line 28:
```suggestion
/// @brief  User function to throw an exception. 
/// > Note: If macro `DISABLE_EXCEPTIONS` is defined, this function terminates printing the message. In this case, the whole code should be considered noexcept.
/// 
```

filippobrizzi (approved):

PR #115, opened 2024-07-03
Subscriber: add option to run callback in a separate thread

filippobrizzi:
# Description
Subscriber: add the option to run callback in a separate thread.
When a message arrives, instead of being executed immediately it is pushed into a queue where a separate worker will fetch and execute it.

Add a new utility class `QueueConsumer` that waits on a queue for new data to be processed and when new data arrives call the user-provided callback on the data.


modules/concurrency/include/hephaestus/concurrency/queue_consumer.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace heph::concurrency {
+
+/// QueueConsumer creates a thread that wait for messages to be pushed in the queue and call the user
+/// provided callback on them.
+template <typename T>
+class QueueConsumer {
+public:
+  using Callback = std::function<void(const T&)>;
+  QueueConsumer(Callback&& callback, std::optional<std::size_t> max_queue_size);
```
johannes-graeter, line 20:
```suggestion
  [[nodiscard]] QueueConsumer(Callback&& callback, std::optional<std::size_t> max_queue_size);
```
perhaps it makes sense to mark the constructor nodiscard, since the thread  is joined when going class goes out of scope?

filippobrizzi: Interesting, I never realized you could mark a constructor nodiscard.
Alternativly we could use the start/stop pattern.

floriantschopp (approved):
lgtm, thanks a lot!

PR #119, opened 2024-07-11
Telemetry

filippobrizzi:
# Description
There are a lot of files, but the most important ones to begin with are:
* `telemetry.h`
  * Contains the functions to log the metrics -> please review the interface
*  `sink.h` 
  * Contains the telemetry sink interface and the data definition
* `example.cpp` 
  * Shows how using telemetry looks like, in particular, it shows how to log metric for types that are
  * Protobuf serializable
  * Json serializable, both custom or via nlohmann::json
  * Single metric 
* `serdes/json.h` 
  * provides utilities to ease conversion to/from json  

`influxdb_sink.cpp` shows how the system can be used to log data to influxdb. In the Readme there are the instruction to test this yourself.


modules/telemetry/examples/example.cpp
```diff
@@ -0,0 +1,80 @@
...
+    heph::telemetry::Telemetry::registerSink(terminal_sink.get());
+    auto rest_sink = heph::telemetry::createRESTSink({ .url = "http://127.0.0.1:5000" });
+    heph::telemetry::Telemetry::registerSink(rest_sink.get());
+    auto influxdb_sink = heph::telemetry::createInfluxDBSink(
+        { .url = "localhost:8087", .token = "my-super-secret-auth-token", .database = "hephaestus" });
+    heph::telemetry::Telemetry::registerSink(influxdb_sink.get());
+
+    auto motor = std::async(std::launch::async, &heph::telemetry::examples::runMotor);
+    auto slam = std::async(std::launch::async, &heph::telemetry::examples::runSLAM);
+
```
lhruby, line 72:
```suggestion
    heph::utils::TerminationBlocker::waitForInterrupt();
```

modules/telemetry/examples/example.cpp
```diff
@@ -0,0 +1,80 @@
...
+    motor.get();
+    slam.get();
+  } catch (std::exception& e) {
+    fmt::println(stderr, "Execution terminated with exception: {}", e.what());
```
lhruby, line 76:
```suggestion
    fmt::println(stderr, "Execution terminated with exception: {}", e.what());
    return EXIT_FAILURE;
```

modules/telemetry/examples/example.cpp
filippobrizzi: not really, there is an infinite loo inside the run functions so it is not correct to wait for interrupt here. We just need to wait for the functions to terminate

modules/telemetry/include/hephaestus/telemetry/telemetry.h
```diff
@@ -0,0 +1,64 @@
...
+
+class Telemetry {
+public:
+  static void registerSink(ITelemetrySink* sink);
+
+  // Define the function input variable
+  template <serdes::JSONSerializable DataT>
+  static void log(const std::string& component, const std::string& tag, const DataT& data,
+                  ClockT::time_point log_timestamp = ClockT::now());
+
```
lhruby, line 22: Given these two are called by the user, add documentation and example here. Looking at example.cpp, things are very clear, so you could just drag this info over.

johannes-graeter:
unasked for review alert ;) sorry to chip in, i thought it made sense, since Filippo and i discussed on discord quite some about this topic.
thanks for the great work! i really like the interface as well. 
however it feels to me like the separation between logs and metrics needs to be sharper. 
since metrics are a lot more valuable when they come at large frequ (like motor speed) this separation is important in my opinion. that being sad every time i have the urge to strong type a log, i  actually want a metric and the logs are usually dynamic in my experience.

also we could perhaps layout the interface for the metrics with the logger now to have a conscise solution?
imo renaming the overload for the jsonSerializable from 'log' to 'measure' would already be cool. we could then use the sinks you defined for now and can later distignuish between metrics sinks (like prometheus_cpp) and logs sinks, since they need different requirements concerning flushing and so on...

johannes-graeter:
I personally also find json hard to read for logs. I personally like logfmt format a bit leaner to read in raw. Since most probably we will use tools like lviz to view them it is more of a sidenote though

filippobrizzi:
> unasked for review alert ;) sorry to chip in, i thought it made sense, since Filippo and i discussed on discord quite some about this topic. thanks for the great work! i really like the interface as well. however it feels to me like the separation between logs and metrics needs to be sharper. since metrics are a lot more valuable when they come at large frequ (like motor speed) this separation is important in my opinion. that being sad every time i have the urge to strong type a log, i actually want a metric and the logs are usually dynamic in my experience.
> 
> also we could perhaps layout the interface for the metrics with the logger now to have a conscise solution? imo renaming the overload for the jsonSerializable from 'log' to 'measure' would already be cool. we could then use the sinks you defined for now and can later distignuish between metrics sinks (like prometheus_cpp) and logs sinks, since they need different requirements concerning flushing and so on...

@johannes-graeter  I realized it is not clear at all: everything you see here is just for `metrics`! I will rename `log` to `metric`.
I still don't know what is the best approach for logs and one open question I have is if it makes actually sense to handles logs and metrics from this class, as we need to be able to produce a `LogEntry` from both (which also makes me think I should probably rename `LogEntry` to `TelemetryEntry`).
Tha being said, does the current design makes more sense?

filippobrizzi:
@johannes-graeter The biggest open question for me regarding metric is the design of `MetricEntry`, and if the two `Telemetry::metric` functions cover all the use cases

johannes-graeter:
Oh yeah that makes sense! Also protobuf is a good choice then since we might wanna publish them...

I have a slight tendency to actually even delete the "general" interface for metrics and only allow typed ones.
Reason:
For metrics you probably want to show them in some graph, since otherwise the datastream is too large to read it raw. Hence you either way need to know beforehand the format in which you publish the metric, hence you can type it... No need for a dynamic interface in that case.

Whats your opinion? Do i miss a usecase for the dynamic interface?

modules/telemetry/include/hephaestus/telemetry/telemetry.h
```diff
@@ -0,0 +1,64 @@
...
+void Telemetry::metric(const std::string& component, const std::string& tag, const std::string& key,
+                       const DataT& value, ClockT::time_point log_timestamp /*= ClockT::now()*/) {
+  const MetricEntry log_entry{
+    .component = component,
+    .tag = tag,
+    .log_timestamp = log_timestamp,
+    .json_values = fmt::format("{{\"{}\": {}}}", key, value),
+  };
+
+  instance().metric(log_entry);
```
johannes-graeter, line 61: this needs to be protected by a lock_guard, right? (like basically all calls in Telemetry except instance()

filippobrizzi: in theory only `Sink::send` needs to be thread safe. We can relax that constraint by moving the lock guard in the metric. Probably it's a good idea given that sink are user defined and we do not want this burden on them.

modules/telemetry/examples/example.cpp
```diff
@@ -34,34 +37,51 @@ struct NavigationLog {
...
 }
 
+struct ControllLog {
+  double error_m;
+  double drift;
+  int frame_rate;
+  int64_t count;
+};
+// NOLINTNEXTLINE
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(ControllLog, error_m, drift, frame_rate, count)
```
johannes-graeter, line 47: oh wow balck magic! love it

modules/telemetry/include/hephaestus/telemetry/sink.h
```diff
@@ -0,0 +1,45 @@
...
+using ClockT = std::chrono::system_clock;
+
+struct MetricEntry {
+  std::string component;
+  std::string tag;
+  ClockT::time_point log_timestamp;
+  std::string json_values;
+};
+[[nodiscard]] auto toJSON(const MetricEntry& log) -> std::string;
+void fromJSON(std::string_view json, MetricEntry& log);
```
johannes-graeter, line 20: do with NLOHMANN MACRO? guess we lose the nodiscard then unfortunately

modules/telemetry/include/hephaestus/telemetry/telemetry.h
```diff
@@ -0,0 +1,47 @@
...
+#include <absl/base/thread_annotations.h>
+
+#include "hephaestus/serdes/json.h"
+#include "hephaestus/telemetry/sink.h"
+
+namespace heph::telemetry {
+
+void registerSink(ITelemetrySink* sink);
+
+void metric(const MetricEntry& log_entry);
```
johannes-graeter, line 18: nit: since its a function I would prefer a verb here.
how about
```suggestion
void measure(const MetricEntry& log_entry);
```

modules/telemetry/include/hephaestus/telemetry/telemetry.h
```diff
@@ -0,0 +1,47 @@
...
+    .log_timestamp = log_timestamp,
+    .json_values = serdes::serializeToJSON(data),
+  };
+
+  metric(log_entry);
+}
+
+template <typename DataT>
+  requires std::is_arithmetic_v<DataT> || std::is_same_v<DataT, std::string>
+void metric(const std::string& component, const std::string& tag, const std::string& key, const DataT& value,
```
johannes-graeter, line 35: shall we keep the generic interface this or delete  it?

modules/telemetry/include/hephaestus/telemetry/sink.h
johannes-graeter: just saw that this is not possible since we need custom conversion for ClockT

modules/telemetry/src/telemetry.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "hephaestus/telemetry/telemetry.h"
+
+#include <mutex>
+#include <vector>
+
+#include "hephaestus/telemetry/sink.h"
+
+namespace heph::telemetry {
+
+class Telemetry {
```
johannes-graeter, line 13: nice that the singleton is private to the translation unit!

modules/telemetry/src/telemetry.cpp
```diff
@@ -0,0 +1,54 @@
...
+  static void registerSink(ITelemetrySink* sink);
+
+  static void metric(const MetricEntry& log_entry);
+
+private:
+  [[nodiscard]] static auto instance() -> Telemetry&;
+
+private:
+  std::mutex sink_mutex_;
+  std::vector<ITelemetrySink*> sinks_ ABSL_GUARDED_BY(sink_mutex_);
```
johannes-graeter, line 24: +1 for the ABSL guard

johannes-graeter, line 24: would it make sense to transfer ownership of the sinks to Telemetry and use a std::unique_ptr here?
seems dangerous to me that the sinks are defined in an unfit spot and go out of scope...

johannes-graeter (approved):
great, I really like it, just a few minors

modules/telemetry/include/hephaestus/telemetry/telemetry.h
filippobrizzi: I am strongly in favour of removing it, but I want to wait for @floriantschopp feedback if he can make a good case for it.

modules/telemetry/src/telemetry.cpp
filippobrizzi: very good question, probably you are right, as the telemetry is a singleton and global its scope is the one of the whole app if we want to avoid weird segfault we want to pass the ownership. I'll update

modules/telemetry/include/hephaestus/telemetry/telemetry.h
filippobrizzi: @lhruby and @floriantschopp are the naming master so I will defer the decision to them.

modules/serdes/include/hephaestus/serdes/json.h
```diff
@@ -59,16 +64,16 @@ auto serializeToJSON(const T& data) -> std::string {
...
+    protobuf::deserializeFromJSON(json, data);
   } else if constexpr (HasJSONDeserialization<T>) {
-    fromJSON(buffer, data);
+    fromJSON(json, data);
   } else if constexpr (HasNlohmannJSONSerialization<T>) {
-    data = nlohmann::json::parse(buffer).get<T>();
+    data = nlohmann::json::parse(json).get<T>();
   } else {
-    data = rfl::json::read<T>(buffer).value();
+    data = rfl::json::read<T>(json).value();
```
johannes-graeter, line 76: Nit:

This is only taste, but I would prefer template specialization over if/else

modules/serdes/include/hephaestus/serdes/json.h
```diff
@@ -53,20 +60,20 @@ auto serializeToJSON(const T& data) -> std::string {
     const nlohmann::json j = data;
     return j.dump();
   } else {
-    static_assert(JSONSerializable<T>, "No serialization to JSON supported");
+    return rfl::json::write(data);
```
johannes-graeter, line 63: Nice!

johannes-graeter (approved):
Relect-cpp really is a good addition to the stack :)

modules/serdes/include/hephaestus/serdes/json.h
johannes-graeter: you mean like 
https://godbolt.org/z/zdhEEzcY1

PR #122, opened 2024-07-15
[fix] stopRequested and more

floriantschopp:
* Fix comment and brief of stopRequested
* Add logging output
* Fix examples

### DIFF
--- a/modules/utils/src/signal_handler.cpp
+++ b/modules/utils/src/signal_handler.cpp
@@ -7,6 +7,8 @@
 #include <csignal>
 #include <mutex>
 
+#include <absl/log/log.h>
+
 namespace heph::utils {
 
 auto TerminationBlocker::stopRequested() -> bool {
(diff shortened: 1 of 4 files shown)


filippobrizzi (approved):

modules/utils/src/signal_handler.cpp
```diff
@@ -33,6 +35,7 @@ auto TerminationBlocker::instance() -> TerminationBlocker& {
 
 auto TerminationBlocker::signalHandler(int /*unused*/) -> void {
   instance().stop_future_ = instance().app_stop_callback_();
+  LOG(INFO) << "Stop requested.";
```
lhruby, line 38: Do we have a strategy for verbose logging in hephaestus? I feel VLOG will come back to us a few more times in the future.

lhruby (approved):

modules/utils/src/signal_handler.cpp
floriantschopp: I would just use VLOG for that, but typically VLOG is rather for stuff that is logged very often, not for one-off logs like this one...

PR #125, opened 2024-07-16
add link markdown for ros comparison and link to readme

johannes-graeter:
# Description

Picking up the discussion on discord I added a comparison to ros readme so that people coming from the ros world can understand the architectural differences hephaestus has and why its good.

## Type of change
- Docs


comparison_to_ros.md
```diff
@@ -0,0 +1,11 @@
+# Comparison to ROS
+
+* We want to move out of ROS approach to define microservices, where every component is a node and we create a massive data flow graph that gets hard to handle.
+* The benefit of that approach is that it is easy to extend and to compose from existing components, however from our experience robotic systems that are past prototype stage do not have that requirement so the cost of it outweighs the benefit
```
filippobrizzi, line 4:
```suggestion
* The benefit of that approach is that it is easy to extend and to compose from existing components, however from our experience robotic systems that are past prototype stage do not have that requirement so the cost of it outweighs the benefit.
```

comparison_to_ros.md
```diff
@@ -0,0 +1,11 @@
...
+* We have a clear idea of what we want: determinism, easy inspectability and efficiency. This can be done with ROS but it is hard to do.
+* We want a standard message format that is not custom but widely used and highly compatible
```
filippobrizzi, line 6:
```suggestion
* We want a standard message format that is not custom but widely used and highly compatible.
```

comparison_to_ros.md
```diff
@@ -0,0 +1,11 @@
...
+
+Example:
```
filippobrizzi, line 8:
```suggestion
*Example*:

```

comparison_to_ros.md
```diff
@@ -0,0 +1,11 @@
...
+For the Visual SLAM module instead of having one feature extraction, a feature cleaning, a prior estimation, a bundle adjustment and a mapping module, in hephaestus this all would be defined in one module that has images as input and map and pose as output.
```
filippobrizzi, line 9: Just one remark, we wouldn't implement SLAM inside hephaestus, but using Hephaestus, so here I would reword as:
```suggestion
For the Visual SLAM module instead of having one feature extraction, a feature cleaning, a prior estimation, a bundle adjustment and a mapping module, all would be defined in one module that has images as input and map and pose as output.
```

comparison_to_ros.md
```diff
@@ -0,0 +1,11 @@
+# Comparison to ROS
```
filippobrizzi, line 1: please create a doc folder and move this there, I am sure we will have more document file in the future

filippobrizzi (commented):
Fantastic!

comparison_to_ros.md
johannes-graeter: done

filippobrizzi (approved):

PR #126, opened 2024-07-22

johannes-graeter:
# Description
Logging interface proposal.

## Original proposal 
https://github.com/olympus-robotics/hephaestus/pull/110

Before defining the logging itself I would fixate the structural logging interface.
Before it was:
```cpp
Log{"my message"}("field1",field1)("field2",field2);
```
I don't like that too much since its somewhat unexpected since it looks like a call succession.

## Proposed
All sensefull operators to me seem binary operators, especially << and | 
For this we need a Field struct:
```cpp
Log{"my message"} << Field{"field1",field1} << Field{"field2",field2};
// or 
Log{"my message"} | Field{"field1",field1} | Field{"field2",field2};
```
both give nice visual seperation and the meaning of the operators is somewhat clear. I would prefer | though because it would look cleaner to me if we need the function log call like
```cpp
log(Info, Log{"my message"} << Field{"field1",field1} << Field{"field2",field2});
// over
log(Info, Log{"my message"} | Field{"field1",field1} | Field{"field2",field2});
```

for convenience I also proposed a string literal like
```cpp
using namespace heph::telemetry::literals; // linter still doesn't like this need to add excpeption
Log{"my message"} | "field1"_f(field1) | "field2"_f(field2});
```

## Others
Other interfaces could be simple add like
```cpp
Log{"my message"}.add("field1",field1).add("field2",field2);
``` 
which I find a little verbose, or variadic
```cpp
Log{"my message",Fields{"field1",field1,"field2",field2});
```
however I like this one not too much since there is worse visual seperation.

## Things to do
- [ ] decide on log function
- [ ] add more data to Log
- [ ] add interface for sinks
- [ ] support std::format style through variadic interface

your opinions are very much appreciated!

## Type of change

- Bug fix (non-breaking change which fixes an issue) (some linting errors deactivated and todos added)


johannes-graeter:
this is a draft based on https://github.com/olympus-robotics/hephaestus/pull/119

modules/telemetry/include/hephaestus/telemetry/struclog.h
```diff
@@ -0,0 +1,106 @@
...
+ *        'level=info file=struclog.h:123 time=2023-12-3T8:52:02+0 message="adding" id=12345
+ * tag="test"'
+ *
+ *        Remark: We would like to use libfmt with quoted formatting as proposed in
+ *                https://github.com/fmtlib/fmt/issues/825 once its included (instead of sstream).
+ *
+ */
+class Log {
+public:
+  explicit Log(const std::string& msg, std::source_location location = std::source_location::current());
```
filippobrizzi, line 43: QQ: would it make sense to support fmt::format for msg? meaning adding support for variadic templates?

johannes-graeter:
I added a draft for the struclog, logging and sink, in this case absl.
unfortunately te diff is all messed up since I based this upon the draft of telemetry metrics. 
I would suggest the following:
A. You look at absl_log_sink*, struclog.* as well as tests and example/log_sink.cpp
B. after applying suggestions I will implement the logging along what @filippobrizzi did for the metrics after my holidays 

please note: 
1. since I am going on vacation tomorrow the state is not where I would like it to be (maybe even that it doesnt compile, the example is just slapped together.). However I thought that for discussion this is better than nothing. 
2. when copmiling in the docker container I got complains in parts of the code that I didnt touch... are the setings for ci different than for default build in the docker container? (i marked them with and @filippo in the code)

modules/telemetry/include/hephaestus/telemetry/struclog.h
```diff
@@ -0,0 +1,145 @@
...
+ *        logs
+ *        'level=info file=struclog.h:123 time=2023-12-3T8:52:02+0 message="adding" id=12345 tag="test"'
+ *
+ *        Remark: We would like to use libfmt with quoted formatting as proposed in
+ *                https://github.com/fmtlib/fmt/issues/825 once its included (instead of sstream).
+ *
+ */
+class LogEntry {
+public:
+  using Fields = std::vector<Field<std::string>>;
```
filippobrizzi, line 45: I would have like for metric a variant here, actually we can use the same variant.

modules/telemetry/examples/log_sink.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+// NOLINTNEXTLINE(bugprone-exception-escape)
+auto main(int /*argc*/, const char* /*argv*/[]) -> int {
+  const heph::utils::StackTrace stack;
+
+  ht::registerLogSink(std::make_unique<ht::AbslLogSink>());
+
+  const std::string a = "testing absl log sink";
+  const int num = 123;
+  ht::log(ht::LogEntry{ ht::Level::Warn, a } | "num"_f(num));
```
filippobrizzi, line 21: This is great!! Love the literals, I guess this works also like?
```suggestion
  ht::log(ht::LogEntry{ ht::Level::Warn, a } | {"num", num});
```
Ideally it would be even more awesome if this would work:

```suggestion
  ht::log({ ht::Level::Warn, a } | "num"_f(num));
```

but I guess C++ is not able to deduce the type?

filippobrizzi: also it would be interesting to start thinking how we could handle `log_if` and `log_every_n` or `log_every_ts` .

modules/telemetry/include/hephaestus/telemetry/struclog.h
johannes-graeter: Finally coming back to this :) To clarify: you would prefer saving the typed values in a variant instead of converting them to string and saving them in the LogEntry? the motivation behind this was to distinguish types that need special rendering in the | operator throught concepts (here Quotable and NonQuotable) and save them all in a format that can be piped to the output sink. I thought that it would anyway be streamed to some string format so rendering them there would remove some resp from the sink implementor...

filippobrizzi: Valid point. In the case of metrics we want variants because the sink will also need to use them as numbers for e.g. influxdb. But in this case you are right, a string should be sufficient as otherwise it should probably be metric.

modules/telemetry/examples/log_sink.cpp
johannes-graeter: added an ticket for this here https://github.com/olympus-robotics/hephaestus/issues/199 to handle in a seperate PR

johannes-graeter: about 
´´´cpp
ht::log({ ht::Level::Warn, a } | "num"_f(num));
```
I have to try around with that, ht::log has only one single defnition as log(const LogEntry&) so it should work as 
´´´cpp
ht::log({ ht::Level::Warn, a });
```
not sure about the piping though

PR #127, opened 2024-07-23
Update zenohcxx and link it correctly.

filippobrizzi:
# Description
Update `zenohcxx` to the latest commit where we fixed the cmake install script.
Now we can properly link it.


floriantschopp (approved):
nice, we are now on a major release :tada:

PR #128, opened 2024-07-23
Enhance JSON serialization

filippobrizzi:
# Description
* Refactor serialization:
  *  Move JSON serialization to a dedicated header
  * Move to template specialization instead of  `if constexpr`
    * This makes the code cleaner
    * Has one important change: a struct cannot support multiple serialization methods -> this ensures no confusion happens when serialization/deserializing.
      * If a struct supports multiple serialization methods the compiler will complain with: `call of overloaded 'serialize(...)' is ambiguous` 
* Add JSON support:
  * with `fromJSON` / `toJSON` user-defined function
  * with `Nlohmann`

### DIFF
--- a/modules/serdes/tests/tests.cpp
+++ b/modules/serdes/tests/tests.cpp
@@ -3,12 +3,15 @@
 //=================================================================================================
 
 #include <cstddef>
+#include <string>
+#include <string_view>
 #include <utility>
 #include <vector>
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "hephaestus/serdes/json.h"
 #include "hephaestus/serdes/protobuf/buffers.h"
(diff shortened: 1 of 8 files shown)


modules/serdes/tests/tests.cpp
```diff
@@ -37,6 +41,25 @@ struct Data {
   int a = NUMBER;
 };
 
+struct DummyJSONSerializable {
+  auto operator==(const DummyJSONSerializable& other) const -> bool = default;
```
lhruby, line 45:
```suggestion
  auto operator==(const DummyJSONSerializable&) const -> bool = default;
```
nit. Can be omitted.

lhruby (approved):
LGTM

floriantschopp (approved):
lgtm

PR #129, opened 2024-07-23
Telemetry measure

filippobrizzi:
# Description
New interface to log measures.
Custom sink implementation will be provided in a following PR.


johannes-graeter (commented):
really like it!

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+namespace heph::telemetry {
+using ClockT = std::chrono::system_clock;
+
+struct MeasureEntry {
+  [[nodiscard]] auto operator==(const MeasureEntry& other) const -> bool = default;
+
+  std::string component;
+  std::string tag;
```
johannes-graeter, line 17: some explanatory comment would be cool here since component and tag are not self explanatory names.

johannes-graeter (approved):

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
```diff
@@ -0,0 +1,29 @@
...
+  ClockT::time_point measure_timestamp;
+  std::string json_values;
+};
+
+class IMeasureSink {
+public:
+  virtual ~IMeasureSink() = default;
+
+  virtual void send(const MeasureEntry& measure_entry) = 0;
```
johannes-graeter, line 26: nit: when the variable type is explanatory like this, I fund it good style to leave the variable name out, since its mostly duplicated info

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
```diff
@@ -0,0 +1,30 @@
...
+
+struct MeasureEntry {
+  [[nodiscard]] auto operator==(const MeasureEntry& other) const -> bool = default;
+
+  std::string component;  ///< The component that is logging the measure, e.g. SLAM, Navigation, etc.
+  std::string tag;        ///< The tag of the measure used to identify who created it, e.g. "front_camera",
+                          ///< "motor1", etc.
+  ClockT::time_point measure_timestamp;
+  std::string json_values;
+};
```
lhruby, line 21: Q: what about ids? How do you associate multiple `MeasureEntries` with each other?
Timestamp works, but it's prone to errors if the timestamps are not 100% identical

modules/telemetry/src/measure.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+private:
+  [[nodiscard]] static auto instance() -> Measure&;
+  void processMeasureEntries(const MeasureEntry& entry);
+
+private:
+  std::mutex sink_mutex_;
+  std::vector<std::unique_ptr<IMeasureSink>> sinks_ ABSL_GUARDED_BY(sink_mutex_);
+  concurrency::MessageQueueConsumer<MeasureEntry> measure_entries_consumer_;
+};
```
lhruby, line 34: The singleton design is very lighweight and elegant. It does however come with the downside of repeating the component name over and over again, with the ability to create different components with typos.

From a user perspective, `Measure` could be a member variable of a class, with `component` either being a parameter of filled at compile time. The call would then be along the lines of `measure_.measure(`tag`, `json_values`)` and the timestamp automatically deduced.

If we fix the `component` at compile time however, we lose the ability to reuse modules and associate them with the module which uses them. Example: ICP can be used by component `SLAM`, but also by component `Perception` or `LoadCarrierDetection`.

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
filippobrizzi: Could please make an example?

lhruby: Example:

Synchronized data is produced with timestamp t0. SLAM consumes this data and produces a world pose with timestamp t0.

Planner module consumes the map and SLAM pose at timestamp t0 and plans a trajectory. 

Both modules output telemetry data, and the backend associates the SLAM telemetry with the planner's telemetry.

I don't suggest to add an id immediately, as this is a hypothetical discussion at this point. The question I would like to raise is how to associate telemetry data logged by different modules?

modules/telemetry/src/measure.cpp
lhruby: Not sure we are fully aligned here yet. If a low level module, like ICP, produces telemetry output, we would like to associate this output via the component with it's caller, e.g. SLAM or LcDetection. The tag would be "icp" in both cases. How would you implement this given the current design, where `record` specifies the component.

One solution would be to add a parent_component member to low level modules, this would however mix the module definition with telemetry specific concerns.

modules/telemetry/tests/data_point_record_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include "hephaestus/telemetry/data_point_record.h"
+#include "hephaestus/telemetry/data_point_sink.h"
+
+// NOLINTNEXTLINE(google-build-using-namespace)
+using namespace ::testing;
+
+namespace heph::telemetry::tests {
+namespace {
+
+class MockMeasureSink final : public IDataPointSink {
```
lhruby, line 31:
```suggestion
class MockProbeSink final : public IDataPointSink {
```

modules/telemetry/src/measure.cpp
filippobrizzi: The ways I see:
* ICP returns detailed information which are recorded by the caller
* Upon creation ICP is passed with the component name which uses to record information
  * This would solve also the logging 
* We make telemetry not singleton and we need to pass the object everywhere (nightmare)

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
filippobrizzi: ok I see, it is a good point. I don't see a downside in having an ID field as a `std::size_t` value. In influxdb it can set as a tag.

johannes-graeter: +1 for the ids

modules/telemetry/src/measure.cpp
lhruby: I would rule out option 2, as it mixes concerns, as discussed above.

Option 1 might be the preferable solution. It means that we need to be a bit more thoughtful on what we want to log, but that is not necessarily a bad thing. By not providing a simple "record' method to be used in any low level module without care, we encourage to not record unneeded information. I would vote for this one, given the options we have.

Regarding option 3, this indeed seems like a chore. 

Feel free to solve this discussion using option 1.

modules/telemetry/include/hephaestus/telemetry/measure_sink.h
lhruby: Ok, let's add an `id` field then. As said, I don't have an immediate use case, but similar to sensor data, or Perfetto tracing, if we want to associate ("synchronize") metrics later, an id is much easier to work with than timestamps.

lhruby (approved):
Great work!

modules/telemetry/src/measure.cpp
filippobrizzi: With the current design if ICP has a function like:
```
auto solve(...) -> Result;
struct Result {
  Pose3 pose;
  Metrics metrics
};
```
in the caller module I can do:
```
auto icp_result = icp->solve(...);
heph::telemetry::record("LIDARSlam", "localization", icp_result.metrics).
```

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+#include <absl/base/thread_annotations.h>
+#include <absl/log/log.h>
+
+#include "hephaestus/concurrency/message_queue_consumer.h"
+#include "hephaestus/telemetry/metric_sink.h"
+
+namespace heph::telemetry {
+
+namespace internal {
```
johannes-graeter, line 26: +1 for internal namespace

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+namespace {
+[[nodiscard]] auto stringToInt64(const std::string& str) -> std::optional<int64_t> {
```
johannes-graeter, line 28: Nit: I personally like to start functions with a verb to indicate action, but more a manner of taste than everything else

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+  const char* start = str.c_str();
+  char* end{};
+  errno = 0;
+
+  static constexpr int BASE = 10;
+  int64_t result = std::strtoll(start, &end, BASE);
```
johannes-graeter, line 34: IMO the error handling of https://en.cppreference.com/w/cpp/utility/from_chars is superior to strtoll and more modern

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+  jsonToValues(json_obj, values, "");
+
+  return values;
+}
+}  // namespace internal
+
+class MetricRecorder {
+public:
+  explicit MetricRecorder();
+  static void registerMetricSink(std::unique_ptr<IMetricSink> sink);
```
johannes-graeter, line 103: Nit: I liked registerSink better, the Metric sematnic is given through the class and the sink type, so registerMetricSink is redundant

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  static void record(const Metric& metric);
+
+private:
+  [[nodiscard]] static auto instance() -> MetricRecorder&;
+  void processMeasureEntries(const Metric& entry);
```
johannes-graeter, line 109: same here, IMO just call it processEntries (also we decided against Measure nomenclature)

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+private:
+  std::mutex sink_mutex_;
+  std::vector<std::unique_ptr<IMetricSink>> sinks_ ABSL_GUARDED_BY(sink_mutex_);
+  concurrency::MessageQueueConsumer<Metric> measure_entries_consumer_;
```
johannes-graeter, line 114:
```suggestion
  concurrency::MessageQueueConsumer<Metric> entries_consumer_;
```

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,152 @@
...
+};
+
+void registerMetricSink(std::unique_ptr<IMetricSink> sink) {
```
johannes-graeter, line 117: also here I would call it just registerSink and overload on the sink type.
IMO
```
registerSink(myMetricSink);
registerSink(myLogSink);
```
is cleaner

modules/telemetry/tests/metric_record_tests.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "hephaestus/random/random_container.h"
+#include "hephaestus/random/random_generator.h"
+#include "hephaestus/random/random_type.h"
+#include "hephaestus/serdes/json.h"
+#include "hephaestus/telemetry/metric_record.h"
+#include "hephaestus/telemetry/metric_sink.h"
+
```
johannes-graeter, line 24: nlohmann as explicit dependency is missing right? shouldn't this be catched by the compiler you did add flags for this right?

johannes-graeter: ahh perhaps it does not work since you only use the macros here....

johannes-graeter (approved):
really like it only a few minor comments

modules/telemetry/src/measure.cpp
johannes-graeter: I have the feeling, I do not fully understand the problem with the component labeling.
Is it that @lhruby wants to track where the function that logs the metric is used?
if yes this could either be solved by structure like @filippobrizzi suggested (the metric is bubbled up to a distinct component and the tag gives the specifics) or automatically with a stacktrace (however this could become costly and we would think how we would do the caching of the stacktrace so we don't evaluate it each time).
However tot me this sounds like it has some overlap with the "tracing" ...

johannes-graeter: In order to avoid typos, we could use StrongTypes instead of strings to describe the component.
that can be converted to String inside of record(). This would be an optional interface, since the component also might be dynamically determined.

modules/telemetry/src/metric_record.cpp
filippobrizzi: good point, however for conversion methods we either omit the source as we can get it from the input or it becomes very verbose, like `convertStringToInt64`, no?

modules/telemetry/src/metric_record.cpp
filippobrizzi: good point, a left over from multiple refactor

modules/telemetry/src/metric_record.cpp
filippobrizzi: I am not sure, I think it is important to explicitly state that the sink for metrics and log are different.

modules/telemetry/tests/metric_record_tests.cpp
filippobrizzi: good point, I think the clang-tidy check is not 100% accurate

modules/telemetry/src/metric_record.cpp
johannes-graeter: yeah I see your point. I personally would prefer consistency over verbosity, also since this is only used in a small scope... but any version is fine by me

modules/telemetry/src/metric_record.cpp
johannes-graeter: I think since they have different interface types its clear, but we can also have more explicit names here.
Better write 5 letters more than not being expressive

modules/telemetry/src/measure.cpp
filippobrizzi: these are all great points. I suggest we get this in and then we consider a refactor when we have more cases
otherwise we risk spinning around forever

johannes-graeter (approved):

modules/telemetry/include/hephaestus/telemetry/metric_record.h
```diff
@@ -0,0 +1,51 @@
...
+#pragma once
+#include <string_view>
+
+#include "hephaestus/serdes/json.h"
+#include "hephaestus/telemetry/metric_sink.h"
+
+namespace heph::telemetry {
+namespace internal {
+[[nodiscard]] auto
+jsonToValuesMap(std::string_view json) -> std::unordered_map<std::string, Metric::ValueType>;
```
floriantschopp, line 14: I know I am late to the discussion, but I am not suuper in favor of the whole `Metric` name, as it is very easily mistaken with metric vs imperial units. So when you just see it, you can mistaken it for sth that has to do with unit transformations etc...

modules/telemetry/include/hephaestus/telemetry/metric_record.h
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/// @brief Register a new data point sink.
+/// For every data point recorded, the sink will be called to consume the data.
+/// There is no limit on the number of sinks supported.
+void registerMetricSink(std::unique_ptr<IMetricSink> sink);
```
floriantschopp, line 20: as you write it here even in the comment, why not

```suggestion
/// @brief Register a new data point sink.
/// For every data point recorded, the sink will be called to consume the data.
/// There is no limit on the number of sinks supported.
void registerDatapointSink(std::unique_ptr<IDatapointSink> sink);
```

modules/telemetry/include/hephaestus/telemetry/metric_record.h
```diff
@@ -0,0 +1,51 @@
...
+
+/// @brief Record a data point.
+/// The data point is forwarded to all registered sinks.
+/// Sinks process the data point in a dedicated thread, this means that this function is non-blocking and
+/// deterministic.
+void record(const Metric& metric);
```
floriantschopp, line 26:
```suggestion
void record(const Datapoint& datapoint);
```

modules/telemetry/src/metric_record.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include <absl/log/log.h>
+
+#include "hephaestus/concurrency/message_queue_consumer.h"
+#include "hephaestus/telemetry/metric_sink.h"
+
+namespace heph::telemetry {
+
+namespace internal {
+namespace {
+[[nodiscard]] auto stringToInt64(const std::string& str) -> std::optional<int64_t> {
```
floriantschopp, line 30: This looks rather error prone (also only indirectly tested)...why not use `std::stoi` / `std::stol`? if the optional is needed, with a try/catch, e.g.
```cpp
[[nodiscard]] auto stringToInt64(const std::string& str) -> std::optional<int64_t> {
   try { 
      return std::stol(str);
    } catch (...) {
      return std::nullopt;
    }
 }    

```

floriantschopp (approved):
scolgtm, great stuff! An example would be nice to have, but I guess without a sink impl, it doesnt make too much sense? Not increadibly happy about the naming, but I am late to the game anyway :sweat_smile:

modules/telemetry/src/metric_record.cpp
filippobrizzi: I have moved the function to string utils and added complete unit tests

PR #130, opened 2024-07-23
Feat/Expose spin counter

lhruby:
# Description
Expose spin counter in `Spinner` class. This e.g. allows to use the counter to assign monotonous ids to data processed during `spinOnce`. A typical example is a `frame_id` or `frame_index`. 

## Type of change

### DIFF
--- a/modules/concurrency/src/spinner.cpp
+++ b/modules/concurrency/src/spinner.cpp
@@ -89,4 +89,8 @@ void Spinner::addStopCallback(std::function<void()>&& callback) {
   stop_callback_ = std::move(callback);
 }
 
+auto Spinner::spinCount() const -> uint64_t {
+  return spin_count_;
+}
+
 }  // namespace heph::concurrency
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -27,8 +27,11 @@ public:
 
   void start();
   auto stop() -> std::future<void>;
-  virtual void spinOnce() = 0;  // Pure virtual function
   void addStopCallback(std::function<void()>&& callback);
+
+  virtual void spinOnce() = 0;  //!< Override this function to define the spinner's behavior.
+
+  [[nodiscard]] auto spinCount() const -> uint64_t;
 
 private:
   void spin();


filippobrizzi (approved):

modules/concurrency/src/spinner.cpp
```diff
@@ -89,4 +89,8 @@ void Spinner::addStopCallback(std::function<void()>&& callback) {
   stop_callback_ = std::move(callback);
 }
 
+auto Spinner::spinCount() const -> uint64_t {
+  return spin_count_;
```
floriantschopp, line 93: its never incremented and the user has no idea that he needs to in `spinOnce`... Should we wrap spinOnce into a `spinOnceParent` (very bad name) sth like
```
void spinOnceParent() {
   ++spin_count_;
   spinOnce();
}
```

floriantschopp (changes requested):
somewhat unrelated but still a bit related: 
Most of the time, I implenent spinOnce to just sleep. Instead of being pure virtual, should we give a default implementation that just sleeps?

modules/concurrency/src/spinner.cpp
floriantschopp: at the very least, we need to leave a comment

filippobrizzi: it is incremented in `spin()` line 64

lhruby: You actually increment it in `spin()`
`    const auto target_timestamp = start_timestamp_ + (++spin_count_) * spin_period_;`

lhruby:
> somewhat unrelated but still a bit related: Most of the time, I implenent spinOnce to just sleep. Instead of being pure virtual, should we give a default implementation that just sleeps?

what does a spinner do which simply sleeps? can you give an example?

filippobrizzi:
> somewhat unrelated but still a bit related: Most of the time, I implenent spinOnce to just sleep. Instead of being pure virtual, should we give a default implementation that just sleeps?

@floriantschopp  What is the use case for having a spinner that just sleep?

Also, I don't know if you saw my message on discord. 
But we should update this design. The proper way would be instead of using inheritance is to use composition and have `Spinner` get `spinOnce` as callback.
If I do this refactor is going to be a problem for you? do you already use Spinner in many places?

modules/concurrency/src/spinner.cpp
floriantschopp: sorry, missed that

floriantschopp:
we can change the interface, thats ok for now

PR #131, opened 2024-07-24
Telemetry sink module & InfluxDB

filippobrizzi:
# Description
Create  a new `telemetry_sink` module separated from main `telemetry` module. This allows not to have to depend on this if not needed when using telemetry.

Add influxdb sink with an example on how to use it and the code to start a backend with influx db and grafana


modules/telemetry_sink/examples/backend/grafana/provisioning/dashboards/Hephaestus/HephaestusDashboard.json
```diff
@@ -0,0 +1,227 @@
+{
```
johannes-graeter, line 1: nice to have the grafana dashboards committed as well as examples

modules/telemetry_sink/src/influxdb_metric_sink.cpp
```diff
@@ -0,0 +1,75 @@
...
+  static constexpr auto URI_FORMAT = "http://{}@{}?db={}";
+  const auto url = fmt::format(URI_FORMAT, config_.token, config_.url, config_.database);
+  LOG(INFO) << fmt::format("Connecting to InfluxDB at {}", url);
+  influxdb_ = influxdb::InfluxDBFactory::Get(url);
+  if (config_.batch_size > 0) {
+    influxdb_->batchOf(config_.batch_size);
+  }
+}
+
+InfluxDBSink::~InfluxDBSink() = default;
```
johannes-graeter, line 63: nit: IMO it is better to define default destructor in the header, so that its clear that we do not want to do custom memory management

johannes-graeter (approved):
Very nice, the design of the sink and the metric feels nice and consistent

modules/telemetry_sink/src/influxdb_metric_sink.cpp
filippobrizzi: unfortunately it needs to be here because `create` complains when generating the unique_ptr that the type is not complete. Very annoying and to be honest I do not really understand the underlying issue but it is a common problem with factory methods.

PR #134, opened 2024-07-25
Feat/Create types module and add dummy type.

lhruby:
# Description
Types defined here are rather abstract and general. Project specific types are out of scope of this module.

### Goals
* Define data types which generalize accross projects.
* Aid testing of modules which requre data to test IO.
* Reduce redundancy accross modules.
* `types` module shall serve as an example on how `types` can be integrated when using Hephaestus.

### Non-goals
* Project specific data types are out of scope of `heph/types`. Data is project specific, thus do not impose conventions here for users of Hephaestus. This translates into e.g. not adding types like 6DOF poses to `heph/types`.
* Serialization is out of scope of `types`. It is performed in separate modules.
* File IO is out of scope of `heph/types`. Save to and load from file is performed in separate modules.

### Requirements
* In general, data structure declarations shall only depend on STL data types and containers.
* Data structure declarations shall not depend on large header-only libraries like Eigen to prevent dependency creep and to keep compile times low.
* Add a method `random` which creates random and possibly unrealistic test data. This data is really random, not a realistic simulation.
* Add `operator==` for testing purposes. In general, adding `operator==() = default` is sufficient.
* Unify testing for all data structures where possible.

## Type of change


modules/types/README.md
```diff
@@ -0,0 +1,21 @@
+# README: types
```
filippobrizzi, line 1: beautiful! perfectly stated!

modules/types/README.md
```diff
@@ -0,0 +1,21 @@
...
+* In general, data structure declarations shall only depend on STL data types and containers.e`)
+* Add a method `random` which creates random and possibly unrealistic test data. This is just random data, not a realistic simulation.
+* Add `operator==` for testing purposes. In general, adding `operator==() = default` is sufficient.
+* Unify testing for all data structures where possible.
+
+### Non-goals
+* Data is project specific, thus we do not impose conventions here for users of Hephaestus. This translates into e.g. not adding types like 6DOF poses to heph/types.
+* Data structure declarations shall not depend on large header-only libraries like Eigen to prevent dependency creep and to keep compile times low.
+* Do not implement Serialization, which is done in a separate module.
+* Do not implement load from file, which is done separately.
```
filippobrizzi, line 21: I would say all of these points are requirements, where the first part is assertive (i.e. do this), and the second part is negational (don't do this). But technically they are not non-goals. A don't requirements is mandatory, while a non-goal is something you do not focus on but it is not forbidden. 
If we want to keep this list separated I would structure it as:
```
## Requirements
### Does

### Do-not
```

modules/types/include/hephaestus/types/dummy_type.h
```diff
@@ -0,0 +1,68 @@
...
+
+  [[nodiscard]] auto operator==(const DummyType&) const -> bool = default;
+
+  [[nodiscard]] static auto random(std::mt19937_64& mt) -> DummyType;
+
+  DummyPrimitivesType dummy_primitives_type{};
+
+  DummyEnum dummy_enum{};
+
+  using TimestampT = std::chrono::time_point<std::chrono::system_clock>;
```
filippobrizzi, line 51: nit. 
```suggestion
  using TimestampT = std::chrono::system_clock::time_point;
```

modules/types/include/hephaestus/types/type_formatting.h
```diff
@@ -0,0 +1,45 @@
...
+
+template <typename T>
+concept IsClockT = requires(T t) {
+  std::is_same_v<T, std::chrono::system_clock> || std::is_same_v<T, std::chrono::steady_clock> ||
+      std::is_same_v<T, std::chrono::high_resolution_clock>;
+};
+
+template <IsClockT T>
+[[nodiscard]] inline auto toString(const std::chrono::time_point<T>& timestamp) -> std::string {
+  return fmt::format("{:%Y-%m-%d %H:%M:%S}", timestamp);
```
filippobrizzi, line 25: probably we also want the decimal part here after seconds, no?

modules/types/include/hephaestus/types/type_formatting.h
```diff
@@ -0,0 +1,45 @@
+//=================================================================================================
+// Copyright (C) 2024 FILICS GMBH
```
johannes-graeter, line 2: guess this should be hephaestus contributors?

lhruby: Great find, updated.

modules/types/README.md
lhruby: Updated the readme. Wdyt?

filippobrizzi (approved):

modules/types/src/dummy_type.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namespace heph::types {
+
+auto DummyPrimitivesType::random(std::mt19937_64& mt) -> DummyPrimitivesType {
+  return { .dummy_bool = heph::random::randomT<bool>(mt),
+           .dummy_int8_t = heph::random::randomT<int8_t>(mt),
+           .dummy_int16_t = heph::random::randomT<int16_t>(mt),
+           .dummy_int32_t = heph::random::randomT<int32_t>(mt),
+           .dummy_int64_t = heph::random::randomT<int64_t>(mt),
+           .dummy_uint8_t = heph::random::randomT<uint8_t>(mt),
```
Fellfalla, line 21: Add note as comment or in description why only adding fewer types, e.g. bool, uint8,  int16, float32 would be insufficient? In my understanding, you wanna add all primitive types to maximize  test coverage.

lhruby: Can you elaborate, which type are you missing?

Fellfalla: What I meant was the other way around. Do you want have **all** types as part of your dummy type? If yes -> you could add a comment for the reader why the list of used types shall be as complete as possible.

Fellfalla:
Code looks good for for making open source w.r.t. IP.

modules/types/src/dummy_type.cpp
lhruby: Gotcha. Added a comment, and added both steady clock and system clock to improve coverage.

PR #136, opened 2024-07-30
Feat/dummy type serialization

lhruby:
# Description
Module **types_protobuf** provides mapping of **types** data structures to proto IDL.

## Detailed description

Protobuf is used to serialize all data. Protobuf files are human readable, and export to JSON is supported out of the box. This allows to dump data into files and read it back later in a human readable format.

### Goals
* proto IDL files for a type are feature complete, i.e. serializing and deserializing leads to the exact same data structure
* provide generic helper functions for serialization (e.g for `std::chrono::time_point`)

### Non-goals
* similar to module **types**, **types_protobuf** is meant to be generic, and project specific (non-generalizable) methods shall not be added

### Requirements
* Types are independent of their serialization method, but the module **types_protobuf** depends on module **types**.

## Type of change


modules/types_protobuf/src/dummy_type.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+//=================================================================================================
+// DummyType
+//=================================================================================================
+auto toProto(proto::DummyType& proto_dummy_type, const DummyType& dummy_type) -> void {
+  toProto(*proto_dummy_type.mutable_dummy_primitives_type(), dummy_type.dummy_primitives_type);
+
+  proto_dummy_type.set_dummyenum(getAsProto(dummy_type.dummy_enum));
+
+  ::heph::types::toProto(*proto_dummy_type.mutable_dummytimestamp(), dummy_type.dummy_timestamp);
```
filippobrizzi, line 131: why do you need `::heph::types`?

filippobrizzi (approved):

modules/types_protobuf/proto/hephaestus/types/proto/dummy_type.proto
```diff
@@ -0,0 +1,47 @@
...
+  float dummyFloat = 10;
+  double dummyDouble = 11;
+}
+
+message DummyType {
+  DummyPrimitivesType dummy_primitives_type = 1;
+
+  DummyTypeDummyEnum dummyEnum = 2;
+
+  google.protobuf.Timestamp dummyTimestamp = 3;
```
floriantschopp, line 43: if we define a timestamp as std::chrono::timepoint, we cannot use `google.protobuf.Timestamp` as a full equivalent due to unbalanced data storage

modules/types_protobuf/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# =================================================================================================
+# Copyright (C) 2023-2024 HEPHAESTUS Contributors
+# =================================================================================================
+
+declare_module(
+  NAME types_protobuf
```
floriantschopp, line 6: internally we went for `types_proto` naming for all proto interface of native `types` module.

modules/types_protobuf/include/hephaestus/types_protobuf/std_chrono_timepoint.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+
+#include <google/protobuf/timestamp.pb.h>
+
+#include "hephaestus/serdes/protobuf/concepts.h"
+
+namespace heph::types {
```
floriantschopp, line 15: we wanted to make sure that we have full control over the clock and not randomly choose steady vs system clock. It has major implications on time synchronization / translation etc. we went for a structure like this:
```
struct Clock {
  using rep = int64_t;
  using period = std::nano;  // nanoseconds
  using duration = std::chrono::duration<rep, period>;
  using time_point = std::chrono::time_point<Clock>;
  static constexpr bool is_steady = true;

  static time_point now() noexcept {
    return time_point(
        duration(std::chrono::steady_clock::now().time_since_epoch().count()));
  }
};

using Timestamp = Clock::time_point;
using Duration = Clock::duration;
```

modules/types_protobuf/tests/serialization_tests.cpp
```diff
@@ -0,0 +1,47 @@
...
+namespace heph::types::tests {
+
+struct TestTag {};
+
+template <class T>
+class SerializationTests : public ::testing::Test {};
+using SerializationImplementations = ::testing::Types<DummyType>;
+TYPED_TEST_SUITE(SerializationTests, SerializationImplementations);
+
+TYPED_TEST(SerializationTests, TestEmptySerialization) {
```
floriantschopp, line 24: as mentioned before, we went the other way around, creating a random proto, which can be done with reflection without the whole random module, and then deserialize it and serialize it again. Should give the same amount of insights (still only one side completeness is guarateed) with way less customized code

modules/types_protobuf/proto/hephaestus/types/proto/dummy_type.proto
lhruby: How come? `google.protobuf.Timestamp` uses two fields internally, `seconds` and `nanoseconds`. That should be sufficient, right?

In the `timestamp` serialisation, there is also a conversion in case `duration` of the chrono clock is below `nanoseconds.

modules/types_protobuf/CMakeLists.txt
lhruby: Can you elaborate? `types_proto` is the name of the `photo` module, `types_protobuf` is the name of the module which contains the conversion `toProto` and `fromProto`. So there are
* `types`
* `types_proto` (somewhat hidden and usually only included in `types_protobuf`)
* `types_protobuf`

modules/types_protobuf/include/hephaestus/types_protobuf/std_chrono_timepoint.h
lhruby: nice!

modules/types_protobuf/tests/serialization_tests.cpp
lhruby: Let's discuss your approach in our meeting. I like the idea. The best code is no code, right?

modules/types_protobuf/src/dummy_type.cpp
lhruby: Not needed, removed.

modules/types_protobuf/proto/hephaestus/types/proto/dummy_type.proto
lhruby: Following our discussion, we conclude that there is no generic solution here. The proposed solution of matching the clock data storage sizes and the clock precision is moved into follow up work.

modules/types_protobuf/CMakeLists.txt
lhruby: Such change in convention shall be discussed further in our next call with @filippobrizzi and @floriantschopp .

modules/types_protobuf/tests/serialization_tests.cpp
lhruby: We concluded that both random in cpp and random in proto are needed to achieve full test coverage for missing fields.

modules/types_protobuf/CMakeLists.txt
floriantschopp: as mentioned, `from/toProto` and the protos themselves can be in the same module, which we call `module_proto`

floriantschopp:
I still don't think the timestamp handling in the PR makes sense. There are two considerations
* *Timestamp*: A timesatmp, aka `std::chrono::time_point` is in theory independent of the clock source, but it has some requirements
  * Accuracy: high enough accuracy for most tasks, imo, this is very well fulfulled by `std::chrono::nanoseconds`
  * Representation: We should definitely not use a floating type here, with `int64_t` we achieve comparability and high range of +- 300 years from 1970. All this while keeping it efficient with 64 bytes.
  * Time frame: A timestamp should be w.r.t.  to a time frame (e.g. UTC, boot, time zone, ...), this is not supported by `std::chrono::time_point` but can be achieved by variable naming or an enum.
* *Clock*: The clock is essential to determine on how the timestamp is set, there are two options here:
  * monotonic and constant rate, aka `steady_clock`: This is essential for any type of sensor fusion or estimation that the intervals between timestamps represent a real duration
  * syncrhonized, aka `system_clock`: A syncrhonized timestamp is as close to the "true" timestamp as possible, however to achieve this, the clock is required to allow for jumps and/or non constant rate
  
Both clock options have their usecase in robotics and its a non-trivial problem on how to solve it. A suggestion:
* Use  a `time_point` with `rep=int64_t` and `period=nanoseconds`, and a Clock that is a `steady_clock` but offset by the `system_clock` at boot to make it "as representative as possible" --> `MyClock`
* Have a "time-keeper" node running which logs the current offset of `MyClock` to the synchronized clock. This data can be later used in post-processing to make the `MyClock` stamps even more representative if needed.
* (optional) For representations that need an updated timestamp at runtime, e.g. for displaying it to the user, we can have a service-call to the time-keeper that gets an updated timestamp

Imo, we should be very careful on what types of the above we want to allow and encourage in the code. Could we just merge this PR without any specific timestamp handling at all and tackle it later on properly? NOTE: `google.protobuf.Timestamp` does not fit the requirements of the timestamp as it has 96 bytes.

modules/types_protobuf/proto/hephaestus/types/proto/dummy_type.proto
lhruby: Following further discussions, we concluded to remove timestamps from hephaestus for now.

modules/types_protobuf/CMakeLists.txt
filippobrizzi: let's go for 
```suggestion
  NAME types_proto
```

modules/types_protobuf/proto/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# =================================================================================================
+# Copyright (C) 2023-2024 HEPHAESTUS Contributors
+# =================================================================================================
+
+set(SOURCES
+    hephaestus/types/proto/dummy_type.proto
+)
+
+define_module_proto_library(NAME types_proto SOURCES ${SOURCES})
```
filippobrizzi, line 9: rule is :`<original_module>_gen_proto`
```suggestion
define_module_proto_library(NAME types_gen_proto SOURCES ${SOURCES})
```

docker/Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt update && \
     pkg-config gpg wget ca-certificates git-lfs curl ccache ninja-build doxygen graphviz linux-generic python3-dev python3-pip iproute2 python-is-python3 \
     net-tools iftop htop lsb-release software-properties-common gnupg bash-completion psmisc less tree && \
     rm -rf /var/lib/apt/lists/*
-RUN python3 -m pip install cmakelang
+RUN python3 -m pip install cmakelang six
```
filippobrizzi, line 13: remove

floriantschopp (approved):

modules/types_protobuf/include/hephaestus/types_protobuf/std_chrono_timepoint.h
lhruby: Let's discuss a proper clock design for Hephaestus in a follow up MR.

docker/Dockerfile
lhruby: Removed.

modules/types_protobuf/proto/CMakeLists.txt
lhruby: Renamed.

Fellfalla (approved):
Looks good to me! Approved.

PR #138, opened 2024-07-31
Action server

filippobrizzi:
# Description
Add an `ActionServer` class, paired with `callActionServer`.

This class is inspired by ROS Action Server and it is basically a `Service` that doesn't return immediately but returns a std::future.

In addition, it provides a functionality to send updates while it's computing.

In the example folder you can find an example server binary and client binary.

Please read the comments in `action_server.h` as they explain some design decisions and limitations.


modules/examples/examples/zenoh_action_server.cpp
```diff
@@ -0,0 +1,93 @@
...
+      std::this_thread::sleep_for(WAIT_FOR);
+    }
+
+    return { accumulated_, iterations_ };
+  }
+
+private:
+  static constexpr auto WAIT_FOR = std::chrono::milliseconds{ 500 };
+  std::size_t accumulated_{};
+  std::size_t iterations_{};
```
lhruby, line 52: What about using the `Accumulator` here?

modules/examples/examples/zenoh_action_server.cpp
```diff
@@ -0,0 +1,93 @@
...
+};
+
+auto main(int argc, const char* argv[]) -> int {
+  const heph::utils::StackTrace stack_trace;
+
+  try {
+    auto desc = getProgramDescription("Binary service client example", ExampleType::ACTION_SERVER);
```
lhruby, line 59: Add a brief note what this example should do.

modules/examples/examples/CMakeLists.txt
```diff
@@ -37,6 +37,20 @@ define_module_example(
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
         PUBLIC_LINK_LIBS hephaestus::ipc)
 
+define_module_example(
+          NAME zenoh_action_server
+          SOURCES zenoh_action_server.cpp zenoh_program_options.h
+          PUBLIC_INCLUDE_PATHS
+          $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../include>
+          PUBLIC_LINK_LIBS hephaestus::ipc)
```
lhruby, line 45: Indentation misalignment.

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,222 @@
...
+#include "hephaestus/ipc/publisher.h"
+#include "hephaestus/ipc/zenoh/internal/action_server_client_helper.h"
+#include "hephaestus/ipc/zenoh/service.h"
+#include "hephaestus/ipc/zenoh/session.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types_protobuf.h"  // NOLINT(misc-include-cleaner)
+
+namespace heph::ipc::zenoh {
+
+/// If an action server is already serving a request it will not accept new ones.
```
lhruby, line 25: It would be great to add a description what an action server does. If you think this is too much common knowledge, please make it brief.

ChatGPT suggestion:
```
In terms of Inter-Process Communication (IPC), an action server is a server that performs specific tasks in response to requests from clients. These tasks are usually well-defined actions that may take some time to complete, such as processing data, performing calculations, or interacting with hardware.

**Key Characteristics of an Action Server:**
**Asynchronous Processing:** Action servers typically handle requests asynchronously, allowing the client to continue its operations while waiting for the action to complete.
**Action Feedback:** While the action is in progress, the server can provide feedback to the client about the status or intermediate results.
**Result Handling:** Once the action is completed, the server sends the final result back to the client.
**State Management:** Action servers often maintain several states such as pending, active, preempted, succeeded, or aborted to manage the lifecycle of the tasks they perform. 
```

modules/examples/include/hephaestus/examples/types/accumulator.h
```diff
@@ -0,0 +1,24 @@
+//=================================================================================================
```
lhruby, line 1: What about Kostya's policy of always having a cpp to include the header, even if its empty?
```
// Contributors header
#include "header"
```

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,222 @@
...
+/// RequestT needs to be copyable and one copy will be made of it.
+/// ActionServer needs two callbacks:
+/// - RequestCallback: to check if the request is valid and if it can be served.
```
filippobrizzi, line 28: Add that it should returns immediately

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,222 @@
...
+/// response `Service` and the update `Publisher` uses a topic name which just depends on the input topic
+/// name. This means that if we have multiple requests we will have multiple `Publisher` and `Subscriber` with
+/// the same topic name, which will cause conflicts. This could be solved by adding a unique identifier to the
+/// topic name, but this is not implemented yet.
+template <typename RequestT, typename StatusT, typename ReplyT>
+class ActionServer {
+public:
+  using RequestCallback = std::function<ActionServerRequestStatus(const RequestT&)>;
+  // NOTE: execution function is noexcept to avoid exceptions in the callback.
+  using ExecuteCallback = std::function<ReplyT(const RequestT&, Publisher<StatusT>&)>;
```
filippobrizzi, line 49: Add also a stop request flag, so the user knows that they should stop the execution

modules/examples/include/hephaestus/examples/types/accumulator.h
filippobrizzi: We actually do, as header files in `examples` or `apps` folder are included in the `.cpp` file of the app. That is needed for library to make sure that somewhere the file is included.

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+/// the same topic name, which will cause conflicts. This could be solved by adding a unique identifier to the
+/// topic name, but this is not implemented yet.
+template <typename RequestT, typename StatusT, typename ReplyT>
+class ActionServer {
+public:
+  using RequestCallback = std::function<ActionServerRequestStatus(const RequestT&)>;
+  using ExecuteCallback =
+      std::function<ReplyT(const RequestT&, Publisher<StatusT>&, std::atomic_bool& stop_requested)>;
+
+  ActionServer(SessionPtr session, TopicConfig topic_config, RequestCallback&& request_cb,
```
filippobrizzi, line 66:
```suggestion
  ActionServer(SessionPtr session, TopicConfig topic_config, RequestCallback&& action_trigger_cb,
```

modules/examples/examples/zenoh_action_server.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "hephaestus/ipc/publisher.h"
+#include "hephaestus/ipc/zenoh/action_server.h"
+#include "hephaestus/ipc/zenoh/session.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types.h"
+#include "hephaestus/utils/signal_handler.h"
+#include "hephaestus/utils/stack_trace.h"
+#include "zenoh_program_options.h"
+
+[[nodiscard]] auto
+request(const heph::examples::types::Accumulator& sample) -> heph::ipc::zenoh::ActionServerRequestStatus {
```
filippobrizzi, line 29:
```suggestion
request(const heph::examples::types::Sample& sample) -> heph::ipc::zenoh::ActionServerRequestStatus {
```

modules/examples/examples/zenoh_action_server_client.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    static constexpr std::size_t START = 42;
+    static constexpr std::size_t ITERATIONS = 10;
+    const auto target = heph::examples::types::Accumulator{
+      .value = START,
+      .counter = ITERATIONS,
+    };
+    auto result_future = heph::ipc::zenoh::callActionServer<heph::examples::types::Accumulator,
+                                                            heph::examples::types::Accumulator,
+                                                            heph::examples::types::Accumulator>(
```
filippobrizzi, line 50: Add different type for request reply

modules/examples/examples/zenoh_action_server.cpp
```diff
@@ -0,0 +1,100 @@
...
+  if (sample.counter == 0) {
+    LOG(ERROR) << "Invalid request, iterations must be greater than 0";
+    return heph::ipc::zenoh::ActionServerRequestStatus::REJECTED_USER;
+  }
+
+  return heph::ipc::zenoh::ActionServerRequestStatus::ACCEPTED;
+}
+
+[[nodiscard]] auto execute(const heph::examples::types::Accumulator& sample,
+                           heph::ipc::Publisher<heph::examples::types::Accumulator>& publisher,
```
filippobrizzi, line 40:
```suggestion
                           heph::ipc::Publisher<heph::examples::types::Accumulator>& status_publisher,
```

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+
+template <typename RequestT, typename StatusT, typename ReplyT>
+void ActionServer<RequestT, StatusT, ReplyT>::execute(const RequestT& request) {
+  is_running_ = true;
+  // NOTE: we create the publisher and the subscriber only if the request is successful.
+  // This has the limit that that some status updates could be lost if the pub/sub are still discovering each
+  // other, but has the great advantage that we do not risk receiving messages from other requests if our
+  // request is rejected. auto publisher = Publisher<StatusT>{ session_,
+  // internal::getStatusPublisherTopic(topic_config_) };
+  auto publisher = Publisher<StatusT>{ session_, internal::getStatusPublisherTopic(topic_config_) };
```
filippobrizzi, line 164:
```suggestion
  auto status_publisher = Publisher<StatusT>{ session_, internal::getStatusPublisherTopic(topic_config_) };
```

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+///   - The updates are decided by the user implementation of the action server.
+/// - Returns a future which will eventually contain the response from the server.
+template <typename RequestT, typename StatusT, typename ReplyT>
+[[nodiscard]] auto callActionServer(SessionPtr session, const TopicConfig& topic_config,
+                                    const RequestT& request, StatusUpdateCallback<StatusT>&& status_update_cb)
+    -> std::future<ActionServerResponse<ReplyT>>;
+
+/// Request the action server to stop.
+[[nodiscard]] auto requestActionServerToStopExecution(Session& session,
+                                                      const TopicConfig& topic_config) -> bool;
```
filippobrizzi, line 105: return `std::atomic_bool`

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+
+// ----------------------------------------------------------------------------------------------------------
```
filippobrizzi, line 107: TODO: add the option to kill the server, to be used only if it gets stuck.

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi, line 66: Add timeout on execute cb

floriantschopp, line 66: we should probably also have a configurable timeout on the request server side?

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+      return execute_cb_(request, publisher, stop_requested);
+    } catch (const std::exception& ex) {
+      LOG(ERROR) << fmt::format("ActionServer (topic: {}): execute callback failed with exception: {}.",
+                                topic_config_.name, ex.what());
+      return ReplyT{};
+    }
+  }();
+  const auto response_topic = internal::getResponseServiceTopic(topic_config_);
+  // TODO: if stop_requested == true, should we communicate somehow that we terminate cause stop was
+  // requested?
```
floriantschopp, line 184: Status could be "Stopped"?

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+  const auto client_response =
+      callService<ReplyT, ActionServerRequestResponse>(*session_, response_topic, reply);
```
floriantschopp, line 186: maybe with timeout?

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+  if (client_response.empty() ||
+      client_response.front().value.status != ActionServerRequestStatus::ACCEPTED) {
```
floriantschopp, line 188: not: shouldn't we check all responses? `std::any_of`?

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  const auto& server_response_status = server_responses.front().value.status;
+  if (server_response_status != ActionServerRequestStatus::ACCEPTED) {
+    return handleFailure<ReplyT>(
+        topic_config.name, fmt::format("request rejected {}", magic_enum::enum_name(server_response_status)),
+        server_response_status);
+  }
+
+  // TODO: figure out why we need this
```
floriantschopp, line 237: what would be the alternative?

modules/ipc/include/hephaestus/ipc/zenoh/internal/action_server_client_helper.h
```diff
@@ -0,0 +1,78 @@
...
+#include "hephaestus/ipc/zenoh/types/action_server_types_protobuf.h"
+
+namespace heph::ipc::zenoh::internal {
+[[nodiscard]] auto getStatusPublisherTopic(const TopicConfig& server_topic) -> TopicConfig;
+
+[[nodiscard]] auto getResponseServiceTopic(const TopicConfig& topic_config) -> TopicConfig;
+
+[[nodiscard]] auto getStopServiceTopic(const TopicConfig& topic_config) -> TopicConfig;
+
+/// If an action server is already serving a request it will reject the new request.
```
floriantschopp, line 26: as mention above, I would add a way to also directly overwrite the running task (with request stop).

modules/ipc/include/hephaestus/ipc/zenoh/types/action_server_types.h
```diff
@@ -0,0 +1,21 @@
...
+//=================================================================================================
+
+#pragma once
+
+#include <cstdint>
+
+namespace heph::ipc::zenoh {
+enum class ActionServerRequestStatus : uint8_t {
+  ACCEPTED = 0,
+  REJECTED_USER = 1,
```
floriantschopp, line 12: acc to above comments, we prob need to add `RESTARTED` and `STOPPED`?

modules/ipc/include/hephaestus/ipc/zenoh/types/action_server_types_protobuf.h
```diff
@@ -0,0 +1,26 @@
...
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#pragma once
+
+#include "hephaestus/ipc/zenoh/proto/action_server_types.pb.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types.h"
+#include "hephaestus/serdes/protobuf/concepts.h"
+
+namespace heph::serdes::protobuf {
```
floriantschopp, line 11: also here, in theory we could auto-generate all that stuff

modules/examples/include/hephaestus/examples/types/accumulator.h
```diff
@@ -0,0 +1,24 @@
...
+};
+}  // namespace heph::examples::types
+
+template <>
+struct fmt::formatter<heph::examples::types::Accumulator> : fmt::formatter<std::string_view> {
+  template <typename FormatContext>
+  auto format(const heph::examples::types::Accumulator& acc, FormatContext& ctx) {
+    return fmt::format_to(ctx.out(), "value: {} | counter: {}", acc.value, acc.counter);
+  }
+};
```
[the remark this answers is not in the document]
filippobrizzi: oh wow that is much better, didn't know about this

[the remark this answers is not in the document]
filippobrizzi: @lhruby  FYI

modules/examples/include/hephaestus/examples/types_protobuf/accumulator.h
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#pragma once
+
+#include "hephaestus/examples/types/accumulator.h"
+#include "hephaestus/examples/types/proto/accumulator.pb.h"
+#include "hephaestus/serdes/protobuf/concepts.h"
+
+namespace heph::serdes::protobuf {
```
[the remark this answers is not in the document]
filippobrizzi: that would be awesome, how would you do it? like `reflect-cpp` can do it for flatbuffer

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: updated, the response needs to be exactly 1, so checking that.

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: the comment got misplaced by the formatter, it was related to the nolint

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,254 @@
...
+#include "hephaestus/ipc/zenoh/session.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types.h"
+#include "hephaestus/ipc/zenoh/types/action_server_types_protobuf.h"  // NOLINT(misc-include-cleaner)
+
+namespace heph::ipc::zenoh {
+
+enum class ActionServerTriggerStatus : uint8_t {
+  SUCCESSFUL = 0,
+  REJECTED = 1,
+};
```
lhruby, line 28:
```suggestion
namspace action_server {
enum class TriggerStatus : uint8_t {
  SUCCESSFUL = 0,
  REJECTED = 1,
};
}
```
nit, style. If you want to avoid a long name which consists of scope + purpose, you could simply place it inside the class with public visibility, or use a namespace.

The proposed solution of ScopePurpose works as well and is the one which I use in Protobuf enums, which always have global visibility. Not my favorite style though.

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,254 @@
...
+                                                      ActionTriggerCallback&& action_trigger_cb,
+                                                      ExecuteCallback&& execute_cb)
+  : session_(std::move((session)))
+  , topic_config_(std::move(topic_config))
+  , action_trigger_cb_(std::move(action_trigger_cb))
+  , execute_cb_(std::move(execute_cb))
+  , request_service_(std::make_unique<Service<RequestT, ActionServerRequestResponse>>(
+        session_, topic_config_, [this](const RequestT& request) { return onRequest(request); }))
+  , request_consumer_([this](const RequestT& request) { return execute(request); }, std::nullopt) {
+  request_consumer_.start();
```
lhruby, line 128: nit. Given the terminology `execute_cb`, should this be called `executor` or similar?

I just read it as request something, i.e. request a consumer first.

modules/ipc/include/hephaestus/ipc/zenoh/types/action_server_types.h
```diff
@@ -0,0 +1,28 @@
+//================================================================================================
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#pragma once
+
+#include <cstdint>
+
+namespace heph::ipc::zenoh {
+enum class ActionServerRequestStatus : uint8_t {
```
lhruby, line 10: similar to above, this compound name can be solved using a namespace `action_server`

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,254 @@
...
+    LOG(ERROR) << fmt::format("ActionServer (topic: {}): server is already serving one request.",
+                              topic_config_.name);
+
+    return { .status = ActionServerRequestStatus::REJECTED_ALREADY_RUNNING };
+  }
+
+  try {
+    const auto response = action_trigger_cb_(request);
+    if (response != ActionServerTriggerStatus::SUCCESSFUL) {
+      return { .status = ActionServerRequestStatus::REJECTED_USER };
```
lhruby, line 149: What is the difference between REJECTED_USER and INVALID?

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,254 @@
...
+  } catch (const std::exception& ex) {
+    LOG(ERROR) << fmt::format("ActionServer (topic: {}): request callback failed with exception: {}.",
+                              topic_config_.name, ex.what());
+    return { .status = ActionServerRequestStatus::INVALID };
+  }
+
+  if (!request_consumer_.queue().tryPush(request)) {
+    // NOTE: this should never happen as the queue is unbound.
+    LOG(ERROR) << fmt::format("ActionServer (topic: {}): failed to push the job in the queue. THIS SHOULD "
+                              "NOT HAPPEN CHECK THE CODE.",
```
lhruby, line 160: I would avoid all caps. A `NOTE: this should not happen ...` should be sufficient.

lhruby (commented):
Great work. 
Two main points
* it would generally be great to break such large feature MRs, where an easy split would be data, logic. Discussions regarding the data will usually revolve around naming, which then does not pollute the logic. Most importantly, the logic is contained which makes feedback easier.
* this might be a matter of style, but I prefer adding namespaces to contain the scope, rather than using scope+purpose names everywhere. This also helps to declutter types

modules/examples/include/hephaestus/examples/types_protobuf/accumulator.h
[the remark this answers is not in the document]
floriantschopp: macro? :monkey_face:

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+
+/// An action server is a server that execute user provided code in response to requests from clients.
+/// Upon completion a result is sent back to the client.
+/// Differently from classic request/response servers, action servers are asynchronous and non-blocking.
+/// Action servers also provide functionalities for the user to send status updates to the client and are
+/// interruptible.
+/// To instantiate an ActionServer the user needs to provide two callbacks:
+/// - RequestCallback
+///   - Takes in input the request and need to decide if the request is valid and if it can be served.
+///   - The callback should not block and return immediately,
```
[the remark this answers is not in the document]
floriantschopp: poke

modules/examples/include/hephaestus/examples/types_protobuf/accumulator.h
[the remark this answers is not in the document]
filippobrizzi: that could be an idea, I am not against that, but you need to run the macro and save the file to git so that other module can have access to the file

modules/ipc/include/hephaestus/ipc/zenoh/internal/action_server_client_helper.h
filippobrizzi: yes I added a todo

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: how do you implement it? We go back to killing the user code

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: great point, I'll add it like for normal `callService`

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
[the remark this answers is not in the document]
filippobrizzi: you are right, rephrasing

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
```diff
@@ -0,0 +1,250 @@
...
+///   - This is the function that does the actual work and eventually return the final response to the client.
+///   - The execute callback is executed in a dedicated thread.
+///   - The function begin execution as soon as the request is accepted.
+///   - The function has access to a Publisher to send status updates to the client.
+///     - The frequency of the updates is decided by the user; updates are not mandatory
+///   - The function can be interrupted by the client by setting the stop_requested flag to true.
+///     - The function should check this flag periodically and return if it is set.
+/// NOTES:
+/// - If an action server is already serving a request it will not accept new ones.
+/// - RequestT needs to be copyable and one copy will be made of it.
```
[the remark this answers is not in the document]
filippobrizzi: right now one thing that is missing is the ability to be notified when the server has completed the request, from a process that is not the one currently waiting for the response. 
We can add such a functionality and then abstract over it.

But first I want to check if it is really needed, because if a server is called only from a single other process than the caller has a future that can use to wait for the process to terminate

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: the problem of putting inside the class is that in this case the class is templated so to use it outside you need to repeat the template, very annoying

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: For example if you call a service with wrong data (of different type) it would be rejected before it reaches the user

filippobrizzi:
> Great work. Two main points
> 
> * it would generally be great to break such large feature MRs, where an easy split would be data, logic. Discussions regarding the data will usually revolve around naming, which then does not pollute the logic. Most importantly, the logic is contained which makes feedback easier.
> * this might be a matter of style, but I prefer adding namespaces to contain the scope, rather than using scope+purpose names everywhere. This also helps to declutter types

I see the struggle, but giving you a PR with just the type in this case wouldn't mean anything to you and you wouldn't be able to give me a meaningful feedback. Unfortunatelly, when introducing new concept the first PR is big. But I tried to keep it compact and have only the bare minimum functionalities. 
I'll give a think about namespace

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
filippobrizzi: usualy function names are verb and class are actor, so executor would be the name if it were a class

floriantschopp (approved):
Great stuff!

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
lhruby: namespace is great

modules/ipc/include/hephaestus/ipc/zenoh/action_server.h
lhruby: good point, leave it

modules/ipc/include/hephaestus/ipc/zenoh/types/action_server_types.h
lhruby: Yes, I would use a folder. Looking at our internal repo, types anyway becomes cluttered quickly, so folders help. AND you solve the issue of name duplication, which naturally occurs in types.

PR #139, opened 2024-08-01
Fix/Rename random module.

lhruby:
# Description
* Unify create random (primitive) types and container types into a single file to improve readability.
* Rename `randomT<T>(...)` into random(...). The namespace `random::random(...)` is strictly needed for correct namespace resolution. This is however only noticable within the `random` module.
* Update filenames accordingly.
* No logic change.

## Type of change


filippobrizzi (approved):

modules/random/include/hephaestus/random/random_object_creator.h
```diff
@@ -0,0 +1,208 @@
...
+  typename T::duration;
+  requires std::is_same_v<typename T::clock, std::chrono::system_clock> ||
+               std::is_same_v<typename T::clock, std::chrono::steady_clock>;
+  requires std::is_same_v<T, typename std::chrono::time_point<typename T::clock, typename T::duration>>;
+};
+
+namespace internal {
+template <IsTimestamp T, size_t Year>
+[[nodiscard]] constexpr auto createFinalTimestampOfTheYear() -> T {
+  // THe final date of the year is YYYY-12-31.
```
floriantschopp, line 82:
```suggestion
  // The final date of the year is YYYY-12-31.
```

modules/random/include/hephaestus/random/random_object_creator.h
```diff
@@ -0,0 +1,208 @@
...
+  constexpr auto YEAR = std::chrono::year{ Year };
+  constexpr auto FINAL_MONTH = std::chrono::December;
+  constexpr auto FINAL_DAY = std::chrono::year_month_day_last{ YEAR / FINAL_MONTH / std::chrono::last };
+  constexpr auto FINAL_DATE_SYS_DAYS = std::chrono::sys_days{ FINAL_DAY };
+
+  // The final time of the day is 23:59:59.000...
```
floriantschopp, line 88: why not 23:59:59.99999? std::chrono is by default afaik in nanoseconds, so why not put sth like
`std:chrono::hours{24} - std::chrono::nanoseconds{1}`

floriantschopp (approved):
in general lgtm, lets have a discussion if we even need the random module

modules/random/include/hephaestus/random/random_object_creator.h
lhruby: That is a good suggestions.
Regarding `nanoseconds`, this is actually more intricate.
* the C++ standard does not specify the precision
* gcc uses `nanoseconds`, that is why linux default is nanosecond, and that is why we assume nanoseconds
* clang uses `microseconds`
* mcvc uses `microseconds`

I can deduce the `clock::duration` here and apply your patch of 24h - our.

PR #143, opened 2024-08-07
Fix: bug in Telemetry; update cmake build tests target name; update CI test step

filippobrizzi:
# Description
* Fix bug in telemetry due to the missed update after the spinner change
* update cmake build tests target name from `tests` to `tests_build`
* update CI from running `check` -> `tests_build` && `test`
 * provide better logging, as it displays progress

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,7 +86,9 @@ jobs:
       run: cmake --build ${{github.workspace}}/build -- all examples
 
     - name: Test
-      # working-directory: ${{github.workspace}}/build
       # Execute tests defined by the CMake configuration.
       # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
-      run: cd ${{github.workspace}}/build && ninja check
+      run: |
+        cd ${{github.workspace}}/build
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -86,7 +86,9 @@ jobs:
...
 
     - name: Test
-      # working-directory: ${{github.workspace}}/build
       # Execute tests defined by the CMake configuration.
       # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
-      run: cd ${{github.workspace}}/build && ninja check
+      run: |
+        cd ${{github.workspace}}/build
+        ninja tests_build
+        ninja test
```
lhruby, line 94:
```suggestion
        ninja tests
        ninja run_tests
```
What about that? we `make tests` to build them, and we `make run_tests` to run them.

lhruby:
LGTM

lhruby (approved):

.github/workflows/build.yml
filippobrizzi: good point, but `test` is a default target generated by cmake so we cannot change that

PR #145, opened 2024-08-09
Fix: Cmake module update add project prefix

filippobrizzi:
# Description
When using Hephaestus as a submodule, if you create a new module with the same name of a module in hephaestus you will get a clash.

This change add internally the project name as a suffix to the module so that we can distinguish.

Minor: fix influxdb name to make it match the module.


lhruby (approved):
LGTM

Issue #147, opened 2024-08-16
[Bag] Flaky test


krisweiss:
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //communication/ipc/heph_ament/hephaestus/modules/bag:tests
-----------------------------------------------------------------------------
Running main() from ./googletest/src/gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Bag
[ RUN      ] Bag.PlayAndRecord
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
I0000 00:00:1723809723.656353      12 zenoh_player.cpp:79] found 2 channels in the bag
I0000 00:00:1723809723.657189      12 zenoh_player.cpp:131] Created publisher for topic: robot
I0000 00:00:1723809723.657699      12 zenoh_player.cpp:131] Created publisher for topic: fleet
W0000 00:00:1723809723.658497      25 zenoh_player.cpp:145] Waiting for subscribers to connect is NOT supported yet!
I0000 00:00:1723809723.659570      23 service.h:154] Calling service on 'type_info/robot'
I0000 00:00:1723809723.661442      22 service.h:128] Received query from 'type_info/robot'
I0000 00:00:1723809723.663892      22 service.h:98] Received answer of 'type_info/robot'
I0000 00:00:1723809723.664045      22 service.h:103] Payload is string: '{"name":"heph.bag.tests.proto.Robot","original_type":"","schema":[10,142,1,10,10,116,101,115,116,46,112,114,111,116,111,18,20,104,101,112,104,46,98,97,103,46,116,101,115,116,115,46,112,114,111,116,111,34,54,10,5,82,111,98,111,116,18,12,10,4,110,97,109,101,24,1,32,1,40,9,18,15,10,7,118,101,114,115,105,111,110,24,2,32,1,40,5,18,14,10,6,115,99,111,114,101,115,24,3,32,3,40,2,34,42,10,5,70,108,101,101,116,18,12,10,4,110,97,109,101,24,1,32,1,40,9,18,19,10,11,114,111,98,111,116,95,99,111,117,110,116,24,2,32,1,40,5,98,6,112,114,111,116,111,51],"serialization":"PROTOBUF"}'
I0000 00:00:1723809723.664973      23 dynamic_subscriber.cpp:84] Create subscriber for topic: robot
[...]

PR #149, opened 2024-08-16
[Fix: bag]: increase timeout to topic database to try to fix bag test

filippobrizzi:
# Description
As reported in this [issue](https://github.com/olympus-robotics/hephaestus/issues/147) the bag test is flaky and can timeout while executing.

We try to increase the timeout on the topic database to give it more time to process the requests.


floriantschopp:
your solution is as good as removing the test…can we increase the timout for the service call?

floriantschopp (approved):

PR #151, opened 2024-08-19
String utils: add toLower; add string to double; fix interface

filippobrizzi:
# Description
Add a couple of new string utilities:
* Extend `stringToInt64` to support also `double` -> `stringTo<T>`
* Add `toLowerCase`
* Update the interface of the other functions to use `std::string_view` instead of `const&`.
  * Why? See this article https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

### DIFF
--- a/modules/utils/src/string/string_utils.cpp
+++ b/modules/utils/src/string/string_utils.cpp
@@ -6,18 +6,14 @@
 
 #include <algorithm>
 #include <cctype>
-#include <charconv>
-#include <cstdint>
 #include <cstdio>
 #include <iterator>
-#include <optional>
 #include <string>
 #include <string_view>
-#include <system_error>
 
 namespace heph::utils::string {
 
-auto toUpperCase(const std::string_view& any_case) -> std::string {
+auto toUpperCase(std::string_view any_case) -> std::string {
   std::string upper_case;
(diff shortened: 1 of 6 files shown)


floriantschopp:
couldn't we use `absl::string_utils` for most of this stuff? eg https://github.com/abseil/abseil-cpp/blob/master/absl/strings/numbers.h

filippobrizzi:
> couldn't we use `absl::string_utils` for most of this stuff? eg https://github.com/abseil/abseil-cpp/blob/master/absl/strings/numbers.h

Great point, I didn't think of checking! Replaced

modules/utils/src/string/string_utils.cpp
```diff
@@ -6,18 +6,14 @@
...
 #include <cstdio>
 #include <iterator>
-#include <optional>
 #include <string>
 #include <string_view>
-#include <system_error>
 
 namespace heph::utils::string {
 
-auto toUpperCase(const std::string_view& any_case) -> std::string {
```
floriantschopp, line 20: also for those we could use https://github.com/abseil/abseil-cpp/blob/master/absl/strings/ascii.h?

floriantschopp (approved):
nice! lgtm

PR #155, opened 2024-08-19
File utils: read/write do not throw anymore; add search for file; add get hostname

filippobrizzi:
# Description
* Read/write file utilities do not throw anymore.
  * In this case it is too aggressive to throw for such a generic function, we need to let the caller decide if the error is at exception level
* Add function to search a file in a list of possible paths
* Add function that return the path of the running executable that calls the function
* Add `getHostname` function: this is used for example to prefix all topics  with the hostname

### DIFF
--- a/modules/utils/src/filesystem/file.cpp
+++ b/modules/utils/src/filesystem/file.cpp
@@ -14,12 +14,15 @@
 #ifdef __linux__
 #include <linux/limits.h>
 #endif
+#include <absl/log/log.h>
+#include <fmt/core.h>
 #include <unistd.h>
 
 namespace heph::utils::filesystem {
 auto readFile(const std::filesystem::path& path) -> std::optional<std::string> {
   std::ifstream infile{ path };
   if (!infile) {
+    LOG(WARNING) << fmt::format("Failed to open file for reading: {}", path.string());
     return {};
   }
 
@@ -33,6 +36,7 @@ auto readFile(const std::filesystem::path& path) -> std::optional<std::string> {
 auto readBinaryFile(const std::filesystem::path& path) -> std::optional<std::vector<std::byte>> {
   std::ifstream infile{ path, std::ios::binary };
   if (!infile) {
+    LOG(WARNING) << fmt::format("Failed to open binary file for reading: {}", path.string());
     return {};
   }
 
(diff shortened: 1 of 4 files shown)


modules/utils/src/filesystem/file.cpp
```diff
@@ -5,20 +5,23 @@
...
+#include <unistd.h>
 
 namespace heph::utils::filesystem {
-auto readFile(const std::filesystem::path& path) -> std::string {
+auto readFile(const std::filesystem::path& path) -> std::optional<std::string> {
   std::ifstream infile{ path };
-  throwExceptionIf<InvalidDataException>(!infile,
-                                         fmt::format("Could not open file {} to read", path.string()));
+  if (!infile) {
+    return {};
```
floriantschopp, line 23: I guess we should add logging here, e.g. `LOG(WARNING)`...

modules/utils/src/filesystem/file.cpp
```diff
@@ -27,10 +30,11 @@ auto readFile(const std::filesystem::path& path) -> std::string {
   return text;
 }
 
-auto readBinaryFile(const std::filesystem::path& path) -> std::vector<std::byte> {
+auto readBinaryFile(const std::filesystem::path& path) -> std::optional<std::vector<std::byte>> {
   std::ifstream infile{ path, std::ios::binary };
-  throwExceptionIf<InvalidDataException>(!infile,
-                                         fmt::format("Could not open file {} to read", path.string()));
+  if (!infile) {
+    return {};
```
floriantschopp, line 36: see above...

modules/utils/src/filesystem/file.cpp
```diff
@@ -40,21 +44,54 @@ auto readBinaryFile(const std::filesystem::path& path) -> std::vector<std::byte>
   return buffer;
 }
 
-void writeStringToFile(const std::filesystem::path& path, std::string_view content) {
+auto writeStringToFile(const std::filesystem::path& path, std::string_view content) -> bool {
   std::ofstream outfile{ path, std::ios::out };
-  throwExceptionIf<InvalidDataException>(!outfile,
-                                         fmt::format("Could not open file {} to write", path.string()));
+  if (!outfile) {
+    return false;
```
floriantschopp, line 50: also for the rest...

floriantschopp (approved):

modules/utils/src/filesystem/file.cpp
filippobrizzi: if we had c++23 we could return `std::expect` with the error message, but I guess for now we need to settle on the warning.

PR #161, opened 2024-09-02
Fix: improve IPC module file structure

filippobrizzi:
# Description
No logic change, just a file restructure with breaking changes:
* Move all the files related to zenoh specifics from `ipc` to `ipc/zenoh` 
* Updated the namespace of the classes moved to zenoh folder to `ipc::zenoh` 
* Renamed publisher subscriber filename to raw_publisher raw_subscriber to match the class name
* Add a new Subscriber template class to match the Publisher
  * This is mainly cosmetics but it is nice to have the type in the subscriber for readability
* Removed from the `examples`  module the ipc command line arguments and use the one defined in the ipc module


modules/ipc/src/zenoh/conversions.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <vector>
+
+#include <fmt/core.h>
+#include <zenoh/api/bytes.hxx>
+#include <zenoh/api/id.hxx>
+#include <zenoh/api/timestamp.hxx>
+
+namespace heph::ipc::zenoh {
+namespace {
+[[nodiscard]] auto toChrono(uint64_t timestamp) -> std::chrono::nanoseconds {
```
lhruby, line 23: Add units to timestamp, or are these `nanoseconds`?

lhruby (approved):
Given you wrote there is no logic change, I assume no code was added, removed or altered, thus approved. If there was a changed, please point to the relevant sections.

floriantschopp (approved):

modules/ipc/src/zenoh/conversions.cpp
filippobrizzi: it is not nanoseconds, it is a weird format described in the documentation linked below

lhruby: merci

PR #169, opened 2024-09-18
Feat/Implement range bounds class.

lhruby:
# Description
* Add a generic class representing range bounds. 
* Add numeric types to `types_proto` to enable serialization of templated classes which are specialized on numeric types.

Usage example: Bounds are often used to represent min and max sensor values, as well as limits on parameters.

## Type of change


filippobrizzi (approved):

modules/types/include/hephaestus/types/bounds.h
```diff
@@ -0,0 +1,100 @@
...
+
+  [[nodiscard]] inline constexpr auto isWithinBounds(T value) const -> bool;
+
+  T lower_bound{};
+  T upper_bound{};
+  BoundsType bounds_type{ BoundsType::INCLUSIVE };
+};
+
+template <NumericType T>
+auto operator<<(std::ostream& os, const Bounds<T>& bounds) -> std::ostream&;
```
floriantschopp, line 42: rather provide a `format_as` instead of this?

modules/types/include/hephaestus/types/bounds.h
```diff
@@ -0,0 +1,100 @@
...
+
+template <NumericType T>
+auto Bounds<T>::random(std::mt19937_64& mt) -> Bounds {
+  return { .lower_bound = random::random<T>(mt),
+           .upper_bound = random::random<T>(mt),
+           .bounds_type = random::random<BoundsType>(mt) };
+}
+
+template <NumericType T>
+inline constexpr auto Bounds<T>::isWithinBounds(T value) const -> bool {
```
floriantschopp, line 56: nit: wouldn't it be nicer to have this as a free function?

modules/types/include/hephaestus/types/bounds.h
```diff
@@ -0,0 +1,100 @@
...
+
+  return os << "Bounds: " << bounds_type_str[0] << bounds.lower_bound << " - " << bounds.upper_bound
+            << bounds_type_str[1] << "\n";
+}
+
+}  // namespace heph::types
+
+namespace fmt {
+template <heph::NumericType T>
+struct formatter<heph::types::Bounds<T>> : ostream_formatter {};
```
floriantschopp, line 99: `fmt::format_as`?

modules/types/include/hephaestus/types/dummy_type.h
```diff
@@ -40,11 +40,11 @@ struct DummyPrimitivesType {
   double dummy_double{};
 };
 
+auto operator<<(std::ostream& os, const DummyPrimitivesType& dummy_primitives_type) -> std::ostream&;
```
floriantschopp, line 43: see above

modules/types_proto/include/hephaestus/types_proto/bounds.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace internal {
+[[nodiscard]] auto getAsProto(const BoundsType& bounds_type) -> proto::BoundsType;
+[[nodiscard]] auto getFromProto(const proto::BoundsType& proto_bounds_type) -> BoundsType;
+}  // namespace internal
+
+template <NumericType T>
+auto toProto(proto::Bounds& proto_bounds, const Bounds<T>& bounds) -> void {
+  toProto(*proto_bounds.mutable_lower_bound(), bounds.lower_bound);
+  toProto(*proto_bounds.mutable_upper_bound(), bounds.upper_bound);
```
floriantschopp, line 38: does this work for built-in types?

modules/types_proto/proto/hephaestus/types/proto/bounds.proto
```diff
@@ -0,0 +1,22 @@
...
+  LEFT_OPEN = 1;   
+  RIGHT_OPEN = 2;  
+  OPEN = 3;         
+}
+
+message Bounds {
+  NumericValue lower_bound = 1;
+  NumericValue upper_bound = 2;
+  BoundsType bounds_type = 3;
+}
```
floriantschopp, line 22:
```suggestion
message Bounds {
enum Type {
  INCLUSIVE = 0;
  LEFT_OPEN = 1;   
  RIGHT_OPEN = 2;  
  OPEN = 3;         
}
  NumericValue lower_bound = 1;
  NumericValue upper_bound = 2;
  BoundsType bounds_type = 3;
}
```
wdyt?

modules/types_proto/src/bounds.cpp
```diff
@@ -0,0 +1,45 @@
...
+      return proto::BoundsType::INCLUSIVE;
+    case BoundsType::LEFT_OPEN:
+      return proto::BoundsType::LEFT_OPEN;
+    case BoundsType::RIGHT_OPEN:
+      return proto::BoundsType::RIGHT_OPEN;
+    case BoundsType::OPEN:
+      return proto::BoundsType::OPEN;
+    default:
+      throwException<InvalidParameterException>("Invalid bounds type");
+      return {};  // unreachable
```
floriantschopp, line 25: one way to get away with not doing that is to unit test the alignment of enums both in proto and c++ types...The unit test can be macro'd and automated. wdyt?

modules/types_proto/src/numeric_value.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include <cstdint>
+
+#include "hephaestus/types/proto/numeric_value.pb.h"
+#include "hephaestus/utils/exception.h"
+
+namespace heph::types::proto {
+
+template <>
+auto toProto(NumericValue& proto_value, int32_t value) -> void {
```
floriantschopp, line 15: not sure if this is the best way to solve this. Wouln't it make more sense to implement a concept that holds true for all types where the proto type has a `.set_value` function? same with `fromProto`?

modules/types_proto/proto/hephaestus/types/proto/bounds.proto
filippobrizzi: if it is posisble to define enums inside messages, yes I agree!

modules/types/include/hephaestus/types/bounds.h
lhruby: Let's see what API we prefer. Currently, we have
```
template <typename T>
class Bounds {
  auto isWithingBounds(T value) -> bool;
};

void main() {
  Bouds<float> current_bounds_amps = { 0.0f, 0.5f };
  float current_amps = 0.3f;
  if(!isWithinBounds(current_amps, current_bounds_amps)) {
    // do something
  }
}
```
your suggestion would mean
```
template <typename T>
class Bounds {
};

template <typename T>
auto isWithinBounds(T value, const Bounds<T>& bounds) -> bool;

Bouds<float> current_bounds_amps = { 0.0f, 0.5f };
float current_amps = 0.3f;
if(!current_bounds_amps.isWithinBounds(current_amps)) {
  // do something
}
```
Give me a vote please @floriantschopp @filippobrizzi

lhruby: I kind of like @floriantschopp suggestion tbh, it reads more natural.

modules/types_proto/include/hephaestus/types_proto/bounds.h
lhruby: Nope, see below.

modules/types_proto/src/numeric_value.cpp
lhruby: I'm torn here. In general, agreed with your statement. 

In this particular case, bounds only make sense for numerical values. Given that this statement is true for other functionality as well, which will be templated on numeric types only, a complete numeric type serialization seems useful.

@filippobrizzi @floriantschopp please wager in here if we want a complete numeric types serialization along the lines proposed here.

modules/types_proto/proto/hephaestus/types/proto/bounds.proto
lhruby: So this code works. However, there are two downsides.
1. I do not want to do the same in our source code, as Bounds is templated, but BoundsType is not, which simplifies things. I would not want to change that.
2. Readability. By removing the explicit enum, protobuf decays all enum values into constants of Bounds, which reduces readability, i.e.
```
case BoundsType::INCLUSIVE:
  return proto::Bounds::INCLUSIVE;
```

Wdyt? Keep the current version or change to the new version? In both cases, I would keep `BoundsType` outside of `Bounds<T>`.

modules/types_proto/src/numeric_value.cpp
filippobrizzi: but for different types there are different functions name, so you cannot really generalize it the only other possible approach is to use `if constexpr` but I think that is less readable

modules/types/include/hephaestus/types/bounds.h
filippobrizzi: let's go free function

modules/types/include/hephaestus/types/bounds.h
filippobrizzi: generally I prefer `format_as` but for some reason most of the times it doesn't work and in this case supporting also ostream make this very compact.

modules/types/include/hephaestus/types/bounds.h
lhruby: Free function it is!

modules/types_proto/proto/hephaestus/types/proto/bounds.proto
lhruby: For readability and linting purposes, we keep enum BoundsType separate.

modules/types_proto/src/numeric_value.cpp
lhruby: We actually need to specialize somewhere, so the current approach is as good as specializing elsewhere. We keep it for now.

modules/types/include/hephaestus/types/bounds.h
lhruby: We keep ostream operator and fmt specialization, as fmt alone would not allow to read google tests output.

floriantschopp: fine with me, formatter is just much harder to read as `format_as` while achieving the same.
For google tests, you can do 
EXPECT_EQ(asdf) << fmt::format(..);

but yes thats no so nice. Alternatively you could define `operator<<` just using `fmt::format` in there.

floriantschopp:
Did you consider having an enfoce method? i.e. something like
```
template<class T>
T Bounds::enforce(const T& t) {
  return std::clamp(t, lower_bound, upper_bound);
}
```
or doing it in-place would also be fine... i.e. `void Bounds::enforce(T& t)` 

wdyt?

modules/types_proto/src/bounds.cpp
lhruby: Maybe we mean different things with aligned. I meant that the underlying int values are the same, which protobuf does not allow AFAIK, given it starts at 0 and monotonously counts up.

In our source code, I for example use a scaler which maps enum values as follows
```
FOO_40_HZ = 1, FOO_20_HZ = 2, FOO_1_HZ = 40
```

If you simply mean that there exists a 1:1 mapping between all enum values, 100% agreed. We'll enforce this soon!

lhruby:
> Did you consider having an enfoce method? i.e. something like
> 
> ```
> template<class T>
> T Bounds::enforce(const T& t) {
>   return std::clamp(t, lower_bound, upper_bound);
> }
> ```
> 
> or doing it in-place would also be fine... i.e. `void Bounds::enforce(T& t)`
> 
> wdyt?

Nice idea, added.

floriantschopp (approved):
lgtm

PR #171, opened 2024-09-20
Feat/Extend type formatting

lhruby:
# Description
* Add enum and std::unordered_map to type formatting
* Extend concepts

## Type of change


modules/utils/include/hephaestus/utils/concepts.h
```diff
@@ -13,6 +13,31 @@ namespace heph {
...
 concept ScalarType = requires(T a) { std::is_scalar_v<T>; };
 
+template <typename T>
+concept EnumType = std::is_enum_v<T>;
+
+template <typename T>
+concept VectorType = requires {
+  typename T::value_type;
+  requires std::same_as<T, std::vector<typename T::value_type>>;
+};
```
filippobrizzi, line 23: maybe we can extend to work also with `std::array`?

modules/types/include/hephaestus/types/type_formatting.h
```diff
@@ -10,28 +10,63 @@
...
+  std::stringstream ss;
+
+  const auto indices = std::views::iota(0);
+  const auto indexed_vec = ranges::views::zip(indices, vec);
 
-template <IsSystemClock T>
+  for (const auto& [index, value] : indexed_vec) {
+    const std::string str = fmt::format("  Index: {}, Value: {}\n", index, value);
+    ss << str;
+  }
```
filippobrizzi, line 33: Maybe let's add a line at the beginning, otherwise if for example you print two vector in a row it is not clear where the first end and the second start (other than seen the index restart).
It could be:
```
fmt::format("[Vector] size: {}, type: {}", vec.size(), utils::getTypeName<T::type>());
```

modules/ipc/apps/zenoh_topic_echo.cpp
```diff
@@ -94,7 +94,7 @@ class TopicEcho {
     truncateLongItems(msg_json, noarr_, max_array_length_);
     fmt::println("From: {}. Topic: {}\nSequence: {} | Timestamp: {}\n{}", metadata.sender_id, metadata.topic,
                  metadata.sequence_id,
-                 types::toString(std::chrono::time_point<std::chrono::system_clock>(
+                 heph::types::toString(std::chrono::time_point<std::chrono::system_clock>(
```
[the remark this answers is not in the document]
lhruby: Indeed, updated.

modules/types/include/hephaestus/types/type_formatting.h
lhruby: `toString` is not to be used in an isolated manner, but simply to parse a generic container. Annotating the result happens in the calling function, see e.g. `DummyType`
```
            << "  dummy_vector=" << toString(dummy_type.dummy_vector) << "\n"
```

modules/utils/include/hephaestus/utils/concepts.h
lhruby: Extended, and added unit tests. PTAL.

filippobrizzi (approved):

PR #174, opened 2024-09-26
Update to Zenoh release 1.0.0.9

filippobrizzi:
# Description
As part of this upgrade, I have:
* Improved the publisher discovery:
  * User callbacks are now called in a separate thread
    * This simplifies `DynamicSubscriber` as it allows the re-use of the same sessions for the discovery and subscribers. 
* Fixed a bug in the publisher where the type info service was created after the creation of the publisher and dynamic subscriber where sometimes failing to get the type info .
* Made bag test more robust by adding prefixes to topics so that when running locally we do not risk subscribing to other topics.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -39,7 +39,9 @@ endif()
 # --------------------------------------------------------------------------------------------------
 # zenoh
 
-set(ZENOHC_CMAKE_ARGS -DZENOHC_CARGO_CHANNEL=+stable -DZENOHC_BUILD_WITH_SHARED_MEMORY=TRUE -DZENOHC_BUILD_WITH_UNSTABLE_API=TRUE)
+set(ZENOHC_CMAKE_ARGS -DZENOHC_CARGO_CHANNEL=+stable -DZENOHC_BUILD_WITH_SHARED_MEMORY=TRUE
+                      -DZENOHC_BUILD_WITH_UNSTABLE_API=TRUE
+)
 if(NOT BUILD_SHARED_LIBS)
   list(APPEND ZENOHC_CMAKE_ARGS "-DZENOHC_LIB_STATIC=TRUE" "-DZENOHC_INSTALL_STATIC_LIBRARY=TRUE")
 endif()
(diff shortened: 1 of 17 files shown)


modules/bag/tests/tests.cpp
```diff
@@ -35,8 +35,8 @@ constexpr auto ROBOT_MSG_PERIOD = std::chrono::milliseconds{ 1 };
 constexpr std::size_t FLEET_MSG_COUNT = 5;
 constexpr auto FLEET_MSG_PERIOD = std::chrono::milliseconds{ 2 };
 constexpr auto SENDER_ID = "bag_tester";
-constexpr auto ROBOT_TOPIC = "robot";
-constexpr auto FLEET_TOPIC = "fleet";
+constexpr auto ROBOT_TOPIC = "bag_test/robot";
+constexpr auto FLEET_TOPIC = "bag_test/fleet";
```
filippobrizzi, line 39: adding a prefix to make sure that if a topic is publishing it doesn't make the test fail

external/CMakeLists.txt
```diff
@@ -39,23 +39,21 @@ endif()
 # --------------------------------------------------------------------------------------------------
 # zenoh
 
-set(ZENOHC_CMAKE_ARGS -DZENOHC_CARGO_CHANNEL=+stable -DZENOHC_BUILD_WITH_SHARED_MEMORY=TRUE)
```
floriantschopp, line 42: do we want the unstable api? what does it enable?

modules/ipc/src/zenoh/dynamic_subscriber.cpp
```diff
@@ -91,12 +88,14 @@ void DynamicSubscriber::onPublisherAdded(const PublisherInfo& info) {
 }
 
 void DynamicSubscriber::onPublisherDropped(const PublisherInfo& info) {
-  throwExceptionIf<InvalidOperationException>(
-      !subscribers_.contains(info.topic),
-      fmt::format("Trying to drop topic {}, but subscriber doesn't exist", info.topic));
+  if (!subscribers_.contains(info.topic)) {
+    LOG(ERROR) << fmt::format("Dropping subscriber for topic: {}, but one doesn't exist", info.topic);
```
floriantschopp, line 92: This description is unclear, what is actually happening?

floriantschopp (approved):

external/CMakeLists.txt
filippobrizzi: cache publisher and being able to get the list of topic and

modules/ipc/src/zenoh/dynamic_subscriber.cpp
filippobrizzi: I will rewrite, but basically we are trying to delete a subscriber which doesn't exists in the first place

PR #176, opened 2024-09-27
Feat/Types proto enum serialization

lhruby:
# Description
Automate serialization of enums using protobuf.

## Type of change


modules/utils/tests/type_formatting_tests.cpp
```diff
@@ -0,0 +1,190 @@
...
+  ASSERT_EQ(str[13], ':');
+  ASSERT_EQ(str[16], ':');
+  ASSERT_EQ(str[19], '.');
+}
+
+<<<<<<<< HEAD:modules/utils/tests/type_formatting_tests.cpp
+}  // namespace heph::utils::string::tests
+========
+}  // namespace heph::utils::format::tests
+>>>>>>>> fix/move-to-string-into-utils-string:modules/utils/tests/format_tests.cpp
```
filippobrizzi, line 190: rebase issue

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
```diff
@@ -0,0 +1,120 @@
...
+
+template <EnumType T, EnumType ProtoT>
+auto toProtoEnum(const T& enum_value) -> ProtoT {
+  static const auto enum_to_proto_enum = internal::createEnumLookupTable<T, ProtoT>();
+  try {
+    return enum_to_proto_enum.at(enum_value);
+  } catch (const std::out_of_range&) {
+    throw std::out_of_range(
+        fmt::format("Enum {} not found in the lookup table", magic_enum::enum_name(enum_value)));
+  }
```
filippobrizzi, line 105:
```suggestion
  throwExceptionIf<InvalidData>(!enum_to_proto_enum.contains(enum_value), "....");
    return enum_to_proto_enum[enum_value];
```

lhruby: Is that efficient though? This code is the one which is being called often, so it is performance critical. I was even thinking about removing the try catch block, as it is only there to extend the exception message by the enum value which is missing.

Wdyt? Remove try catch here alltogether?

modules/utils/tests/type_formatting_tests.cpp
lhruby: removed

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
filippobrizzi: I mean if we really want to make it fast, we should do
```suggestion
  auto enum_it = enum_to_proto_enum.find(enum_value);
  throwExceptionIf<InvalidData>(enum_it == enum_to_proto_enum.end(), "....");
  return enum_it->second;
```

modules/types_proto/proto/hephaestus/types/proto/dummy_type.proto
```diff
@@ -6,7 +6,7 @@ syntax = "proto3";
 
 package heph.types.proto;
 
-enum DummyTypeDummyEnum {
+enum DummyTypeExternalDummyEnum {
   A = 0;
   B = 1;
   C = 2;
```
filippobrizzi, line 12: To follow `buf` requirement, as proto enum are not class, it forces you to append the class name as prefix:
```suggestion
enum DummyTypeExternalDummyEnum {
  DUMMY_TYPE_EXTERNAL_DUMMY_ENUM_A = 0;
  B = 1;
  C = 2;
```
see https://github.com/olympus-robotics/hephaestus/pull/162/files#diff-e47d59905961e3347b18223422fce7ea857673877a9e75a4ffa52d0da6c2bee7R8

filippobrizzi: We will enforce this once we introduce buff, hopefully soon. So I think we can start by now updating the enum names.
You have that nice camel case to scream case function so it should be easy to add the prefix.

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
```diff
@@ -0,0 +1,120 @@
...
+template <EnumType T, EnumType ProtoT>
+auto fromProto(const ProtoT& proto_enum_value, T& enum_value) -> void {
+  static const auto proto_enum_value_to_enum =
+      internal::createInverseLookupTable(internal::createEnumLookupTable<T, ProtoT>());
+  try {
+    enum_value = proto_enum_value_to_enum.at(proto_enum_value);
+  } catch (const std::out_of_range&) {
+    throw std::out_of_range(
+        fmt::format("Proto enum {} not found in the lookup table", magic_enum::enum_name(proto_enum_value)));
+  }
```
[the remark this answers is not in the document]
lhruby: find is actually not what we want. It searches for a value. We want to access the value by providing the key.
I think .at() is good enough. It's safe access, and unless we screw up ,it should never trigger

modules/types_proto/proto/hephaestus/types/proto/dummy_type.proto
lhruby: Ok. But are you sure about the case? If it is a nested enum, proto will create DummyType_InternalDummyEnum_A automatically.

filippobrizzi: Yes, I didn't try for nested enums but for external one yes. I mean we can leave it as it is now and then when I finish the buf PR I will update your prefix function

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
```diff
@@ -0,0 +1,110 @@
...
+#include "hephaestus/utils/string/string_literal.h"
+#include "hephaestus/utils/string/string_utils.h"
+
+namespace heph::serdes::protobuf {
+
+template <EnumType ProtoT, EnumType T>
+[[nodiscard]] auto toProtoEnum(const T& enum_value) -> ProtoT;
+
+template <EnumType ProtoT, EnumType T>
+auto fromProto(const ProtoT& proto_enum_value, T& enum_value) -> void;
```
filippobrizzi, line 26: I wonder if it is not better to be consistent with the `toProtoEnum` and have `fromProtoEnum` and make it return the value.

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  return proto_enum_to_enum;
+}
+}  // namespace internal
+
+template <EnumType ProtoT, EnumType T>
+auto toProtoEnum(const T& enum_value) -> ProtoT {
+  static const auto enum_to_proto_enum = internal::createEnumLookupTable<ProtoT, T>();
+  return enum_to_proto_enum.at(enum_value);
```
filippobrizzi, line 100: you actually need to use my suggestion as we do not want to blindly use exception, for example `throwException` can be set to panic. In Flo usecase they forbid the usage of exception.

modules/types_proto/proto/hephaestus/types/proto/dummy_type.proto
lhruby: Let's do that change at once. We will also have to update forkify, so let's do that in one push and leave the current implementation.

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
lhruby: Your call, both is fine with me. You will however have to specify the non proto type as a template parameter then.

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
lhruby: Ah, great input. I'll refactor then.

modules/serdes/include/hephaestus/serdes/protobuf/enums.h
filippobrizzi: ahh right...ok let's leave it as it is

filippobrizzi (approved):

modules/types_proto/proto/hephaestus/types/proto/dummy_type.proto
lhruby: I just realized that this is only a change in the proto files, not in the cpp enums. I implemented your suggestion in this MR.

PR #180, opened 2024-10-01
Feat/Add limits to random module.

lhruby:
# Description
* Allow to specify limits for certain methods of the `random` module.
Note that `Bounds<T>` cannot be used here, as the module `types` depends on the module `random`, which would lead to cyclic dependencies.

## Type of change

### DIFF
--- a/modules/random/include/hephaestus/random/random_object_creator.h
+++ b/modules/random/include/hephaestus/random/random_object_creator.h
@@ -16,6 +16,7 @@
 
 #include "hephaestus/utils/concepts.h"
 #include "hephaestus/utils/exception.h"
+#include "hephaestus/types/bounds.h"
 
 namespace heph::random {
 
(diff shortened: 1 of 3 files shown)


modules/random/include/hephaestus/random/random_object_creator.h
```diff
@@ -16,6 +16,7 @@
 
 #include "hephaestus/utils/concepts.h"
 #include "hephaestus/utils/exception.h"
+#include "hephaestus/types/bounds.h"
```
lhruby, line 19: @filippobrizzi there is an issue with our current module structure, which leads to cyclic dependencies. 

If we use `types::DummyType` to test other modules, then non of these modules must depend on types. If we add helper types like `Bounds<T>` to `types`, we cannot use them in most hephaestus modules. The random module is a good example.

Solutions: we either move DummyType into a separate module, like a test helper, or we try to formulate all helper types in terms of utility etc. The latter sounds more appealing.

`utility::???::Bounds<T>` <-- suggestions?

filippobrizzi: makes sense, actually `Bounds` is not really a type, it is a helper/utils I would say.

lhruby: Any suggestion for a folder/namespace for `Bounds`? utils::types, utils::data, utils::ranges, ...

filippobrizzi: I vote for `data`

filippobrizzi (approved):

PR #182, opened 2024-10-07
Upgrade to clang 19

filippobrizzi:
# Description
Upgrade to clang 19 and fix new errors, no major change, just following compiler suggestions. 
Some changes also came from making it compile for GCC.

* Fix `zenohd` version installed in docker to match the one from the library.
* Remove dependency docker version, to simplify and reduce the chance of errors that often happen from forgetting to increment the number when upgrading a dependency.
* Add a CMake prefix `gcc` to allow to easily build with GCC.
* To make the test pass for GCC I changed the resolution of the formatted for `time_point` to microseconds as GCC and clang have different default resolution.

### DIFF
--- a/modules/ipc/src/zenoh/scout.cpp
+++ b/modules/ipc/src/zenoh/scout.cpp
@@ -33,8 +33,8 @@ namespace heph::ipc::zenoh {
 
 namespace {
 
-[[nodiscard]] inline auto toStringVector(const std::vector<std::string_view>& str_vec)
-    -> std::vector<std::string> {
+[[nodiscard]] inline auto
+toStringVector(const std::vector<std::string_view>& str_vec) -> std::vector<std::string> {
   std::vector<std::string> output_vec;
   output_vec.reserve(str_vec.size());
   for (const auto& str : str_vec) {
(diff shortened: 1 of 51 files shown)


lhruby (approved):
LGTM. Nice work.

modules/ipc/src/zenoh/scout.cpp
```diff
@@ -103,8 +103,7 @@ auto getListOfNodes() -> std::vector<NodeInfo> {
   ScoutDataManager manager;
 
   auto config = ::zenoh::Config::create_default();
-  ::zenoh::scout(
-      std::move(config), [&manager](const auto& hello) { manager.onDiscovery(hello); }, []() {});
+  ::zenoh::scout(std::move(config), [&manager](const auto& hello) { manager.onDiscovery(hello); }, []() {});
```
floriantschopp, line 106: seems liks sth with the max line length changed? this seems to be a veery long line, no?

floriantschopp (approved):

modules/ipc/src/zenoh/scout.cpp
filippobrizzi: we have a limit of 110 and that line is 109 😅

PR #187, opened 2024-10-21
Fix/Correctly forward exceptions thrown in Spinner callback and allow for stopping

lhruby:
# Description
* Correctly forward exceptions thrown in `spin()`. Since `spin()` runs in a separate thread, we need to propagate the exception message out of `spin`.
* Exception messages are caught during `stop()`, which is usually preceded by a call to `wait()`. `wait()` will exit if an exception is thrown, such that it can be caught in the main thread via `stop()` right away.
* Allow to add a return value to the callback to allow to stop the spinner at completion, i.e. add a `StoppableCallback`
* 
## Type of change
- Bug fix
- New feature

### DIFF
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -30,6 +30,7 @@ public:
 
   void start();
   auto stop() -> std::future<void>;
+  void waitForCompletion();
 
   [[nodiscard]] auto spinCount() const -> uint64_t;
 
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -41,7 +41,6 @@ private:
 private:
   Callback callback_;
 
-  std::atomic_bool is_started_ = false;
   std::atomic_bool stop_requested_ = false;
   std::future<void> async_spinner_handle_;
   std::promise<SpinResult> spin_result_promise_;
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -30,7 +30,7 @@ public:
 
   void start();
   auto stop() -> std::future<void>;
-  void waitForCompletion();
+  void wait();
 
   [[nodiscard]] auto spinCount() const -> uint64_t;
 
@@ -44,7 +44,7 @@ private:
   std::atomic_bool is_started_ = false;
   std::atomic_bool stop_requested_ = false;
   std::future<void> async_spinner_handle_;
-  std::promise<void> exception_thrown_in_spin_promise_;
+  std::promise<SpinResult> spin_result_promise_;
 
   std::chrono::microseconds spin_period_;
   std::chrono::system_clock::time_point start_timestamp_;
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -43,6 +43,7 @@ private:
   std::atomic_bool is_started_ = false;
   std::atomic_bool stop_requested_ = false;
   std::future<void> async_spinner_handle_;
+  std::promise<void> exception_thrown_in_spin_promise_;
 
   std::chrono::microseconds spin_period_;
   std::chrono::system_clock::time_point start_timestamp_;
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -42,7 +42,7 @@ private:
 
   std::atomic_bool is_started_ = false;
   std::atomic_bool stop_requested_ = false;
-  std::thread spinner_thread_;
+  std::future<void> async_spinner_handle_;
 
   std::chrono::microseconds spin_period_;
   std::chrono::system_clock::time_point start_timestamp_;
--- a/modules/concurrency/include/hephaestus/concurrency/spinner.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner.h
@@ -18,7 +18,8 @@ namespace heph::concurrency {
 /// the given fixed rate.
 class Spinner {
 public:
-  using Callback = std::function<void()>;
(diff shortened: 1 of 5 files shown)


modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -29,6 +30,7 @@ class Spinner {
 
   void start();
   auto stop() -> std::future<void>;
+  void waitForCompletion();
```
filippobrizzi, line 33: nit. this should be called `wait` as we use `wait` elsewhere

modules/concurrency/src/spinner.cpp
```diff
@@ -48,8 +48,14 @@ Spinner::~Spinner() {
...
 
-  // NOTE: Replace with std::stop_token and std::jthread when clang supports it.
-  spinner_thread_ = std::thread([this]() { spin(); });
+  async_spinner_handle_ = std::async(std::launch::async, [this]() mutable {
+    try {
+      spin();
+      this->exception_thrown_in_spin_promise_.set_value();
+    } catch (const std::exception& e) {
+      this->exception_thrown_in_spin_promise_.set_exception(std::current_exception());
+    }
```
filippobrizzi, line 57: you do not need this, the exception is automatically captured and returned in the future. 
When you call `async_spinner_handle_.get()` if an exception was raised inside, it will be raised again

modules/concurrency/src/spinner.cpp
```diff
@@ -39,7 +39,7 @@ Spinner::Spinner(Callback&& callback, double rate_hz /*= 0*/)
 }
 
 Spinner::~Spinner() {
-  if (is_started_.load() || spinner_thread_.joinable()) {
+  if (is_started_.load()) {
```
filippobrizzi, line 42: we can remove `is_started_` and replace this with:
```suggestion
  if (async_spinner_handle_.valid()) {
```

modules/concurrency/src/spinner.cpp
```diff
@@ -79,12 +89,25 @@ auto Spinner::stop() -> std::future<void> {
...
+
+  if (async_spinner_handle_.valid()) {
+    try {
+      exception_thrown_in_spin_promise_.get_future().get();  // Check for any exception thrown during spin().
+    } catch (const std::exception& e) {
+      throw e;  // Re-throw the exception to be handled by the caller.
+    }
+  }
+
+  return stop_future;
```
filippobrizzi, line 102: I think we can replace all of this with just 
```suggestion
  return std::move(async_spinner_handle_);
```
and remove `stopImpl`

modules/concurrency/src/spinner.cpp
```diff
@@ -79,12 +89,25 @@ auto Spinner::stop() -> std::future<void> {
...
 }
 
-void Spinner::stopImpl() {
-  spinner_thread_.join();
+void Spinner::waitForCompletion() {
+  async_spinner_handle_.wait();
+  stop().get();
```
filippobrizzi, line 107: wait should not get, the behavior should be the same as future. If you call `wait` you are still supposed to call `stop`
```suggestion
  async_spinner_handle_.wait();
```

modules/concurrency/tests/spinner_tests.cpp
```diff
@@ -12,8 +14,40 @@
 
 namespace heph::concurrency::tests {
 
+struct TestFixture : public ::testing::Test {
+  static auto trivialCallback() -> std::function<Spinner::SpinResult()> {
+    return []() { return Spinner::SpinResult::Continue; };
+  }
+
+  static auto stoppingCallback(size_t& callback_called_counter) -> std::function<Spinner::SpinResult()> {
```
filippobrizzi, line 22:
```suggestion
  static auto stoppingCallback(size_t& callback_called_counter) -> Spinner::Callback {
```

modules/concurrency/tests/spinner_tests.cpp
```diff
@@ -12,8 +14,40 @@
 
 namespace heph::concurrency::tests {
 
+struct TestFixture : public ::testing::Test {
+  static auto trivialCallback() -> std::function<Spinner::SpinResult()> {
```
filippobrizzi, line 18:
```suggestion
  static auto trivialCallback() -> Spinner::Callback {
```

modules/concurrency/tests/spinner_tests.cpp
```diff
@@ -12,8 +14,40 @@
...
+      if (callback_called_counter < STOP_AFTER) {
+        ++callback_called_counter;
+        return Spinner::SpinResult::Continue;
+      }
+
+      return Spinner::SpinResult::Stop;
+    };
+  }
+
+  static auto nonThrowingCallback(size_t& callback_called_counter) -> std::function<Spinner::SpinResult()> {
```
filippobrizzi, line 34:
```suggestion
  static auto nonThrowingCallback(size_t& callback_called_counter) -> Spinner::Callback {
```

modules/concurrency/tests/spinner_tests.cpp
```diff
@@ -12,8 +14,40 @@
...
+    return [&callback_called_counter]() {
+      ++callback_called_counter;
+      return Spinner::SpinResult::Continue;
+    };
+  }
+
+  static auto throwingCallback() -> std::function<Spinner::SpinResult()> {
```
filippobrizzi, line 41:
```suggestion
  static auto throwingCallback() -> Spinner::Callback {
```

modules/concurrency/include/hephaestus/concurrency/spinner.h
lhruby: Renamed.

modules/concurrency/src/spinner.cpp
lhruby: This would work, if we would not move `async_spinner_handle_` in `stop()`. Since we do, that change would seg fault. We need the helper variable `is_started_` since we cannot know if `async_spinner_handle_` has been moved.

filippobrizzi: actually, it is legal to call `valid` on moved object according to documentation https://en.cppreference.com/w/cpp/thread/future/valid

filippobrizzi (approved):

PR #192, opened 2024-10-25
[CMake] Add option to specify a dependency is test only

filippobrizzi:
# Description
To improve modularity and reduce dependencies, I added an option to specify that a module is a test dependency only.
In this way, when building with `BUILD_TESTING=OFF` we can speed things up and reduce coupling.

To showcase the improvement I updated the unit test of the `bag` module:
* Remove the custom proto and instead use `dummy_type` 
* Make `types_proto` and `random` test only dependency


floriantschopp (approved):
niice

PR #193, opened 2024-10-25
Start move Zenoh tests from `examples` to `ipc` modules.

filippobrizzi:
# Description
Now that we can use `types_proto` in test we can better organize tests.
* Moved pub/sub tests 
* Moved service tests
* Add option to random string generator to create characters only strings
  * Cannot create topic with not character strings 

Next step:
* Create Action Server test
* Move type check test


modules/ipc/CMakeLists.txt
```diff
@@ -4,7 +4,15 @@
...
 declare_module(
   NAME ipc
-  DEPENDS_ON_MODULES cli concurrency containers utils serdes
+  DEPENDS_ON_MODULES
+    cli
+    concurrency
+    containers
+    utils
+    serdes
+    DEPENDS_ON_MODULES_FOR_TESTING
```
floriantschopp, line 13:
```suggestion
 DEPENDS_ON_MODULES_FOR_TESTING
```

floriantschopp (approved):

PR #195, opened 2024-10-25
[IPC/Zenoh] add type info to message metadata for publisher

filippobrizzi:
# Description
Zenoh allows to attach metadata to every message, we already use it to share the sequence number and session id. Now it also pass the type info. 

In the subscriber and Service we use it check that the types match, in this way we can better catch the issue.

### DIFF
--- a/modules/bag/tests/tests.cpp
+++ b/modules/bag/tests/tests.cpp
@@ -25,9 +25,13 @@
 #include "hephaestus/types/dummy_type.h"
 #include "hephaestus/types_proto/dummy_type.h"  // NOLINT(misc-include-cleaner)
 #include "hephaestus/utils/filesystem/scoped_path.h"
+<<<<<<< HEAD
+=======
+#include "hephaestus/utils/utils.h"
+    >>>>>>> 168fd6d (squash)
 
-// NOLINTNEXTLINE(google-build-using-namespace)
-using namespace ::testing;
+    // NOLINTNEXTLINE(google-build-using-namespace)
+    using namespace ::testing;
 
 namespace heph::bag::tests {
 namespace {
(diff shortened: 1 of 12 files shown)


floriantschopp (approved):

modules/random/include/hephaestus/random/random_object_creator.h
```diff
@@ -153,12 +153,19 @@ namespace internal {
 /// Generate a random string of characters, including special case characters and numbers.
 template <StringType T>
 [[nodiscard]] auto random(std::mt19937_64& mt, std::optional<size_t> fixed_size = std::nullopt,
-                          bool allow_empty = true) -> T {
+                          bool allow_empty = true, bool lower_characters_only = false) -> T {
```
lhruby, line 156: Options for string creation start to sum up. What about adding a `RandomStringConfig` to bundle them?

lhruby (approved):
lgtm

modules/random/include/hephaestus/random/random_object_creator.h
filippobrizzi: in this way it is possible to have default values

lhruby:
```
struct RandomStringConfig {
  std::optional<size_t> fixed_size = std::nullopt,
  bool allow_empty = true;
  bool lower_characters_only = false;
} RANDOM_STRING_CONFIG_DEFAULT; 

[[nodiscard]] auto random(std::mt19937_64& mt, RandomStringConfig config = RANDOM_STRING_CONFIG_DEFAULT) -> T {
```
The issue with default values is that you can only omit arguments in order, so you can e.g. not say `random(mt, use_only_lower_case_characters)`, so default values are limited. A config does not make things worse.

lhruby: By the way, this should be `lower_case_characters_only` or `force_lower_case_characters`.

filippobrizzi: I would resolve this, and address this in a follow up when we may want to add more options

PR #196, opened 2024-10-25
[IPC/Zenoh] Service and ActionServer add type check

filippobrizzi:
# Description
Add explicit type check for Service and ActionServer requests.

`Service` now also accept a failure callback that invokes if there are errors in the request (this is needed by the action server to unblockitself).


floriantschopp (approved):

lhruby (approved):
lgtm

PR #198, opened 2024-10-31
Add Spinner manager

filippobrizzi:
# Description
The `SpinnersManager` allows to orchestrate the start and termination of multiple spinners. 

The most important functionality is the `waitAny` function which allows to unblock as soon as one of the spinner terminates. In this way we can proceed to stop all of the other. This is very useful in the context of errors.

### DIFF
--- a/modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
@@ -29,7 +29,7 @@ public:
   /// propagated as the other spinners are still blocking.
   void waitAll();
   /// @brief wait until any spinner terminates or throws an exception, which allows for immediate exception
-  /// handling. Not that the exception will be re-thrown when calling `stopAll`.
+  /// handling. Note that the exceptions will be re-thrown when calling `stopAll`.
   void waitAny();
   void stopAll();
 
--- a/modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
@@ -36,7 +36,7 @@ public:
 private:
   std::vector<Spinner*> spinners_;
 
-  std::vector<std::future<void>> wait_futures_;
+  std::atomic_flag termination_flag_ = ATOMIC_FLAG_INIT;
 };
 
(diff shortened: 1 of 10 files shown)


modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
```diff
@@ -0,0 +1,38 @@
...
+/// NOTE: this logic is quite generic and can be extended to any type of object that has `wait()` and `stop()`
+/// methods. To be faitful to the principle of implement only what you need now, we limit the scope to
+/// spinners and consider to expand the scope when an use case arises.
+class SpinnersManager {
+public:
+  explicit SpinnersManager(std::vector<Spinner*> spinners);
+
+  void startAll();
+  void waitAll();
+  void waitAny();
```
lhruby, line 29:
```suggestion
  /// @brief blocks until all spinners are finished. If a single spinner throws an exception, it will not be propagated as the other spinners are still blocking.
  void waitAll();
  /// @brief wait until any spinner terminates or throws an exception, which allows for immediate exception handling.
  void waitAny();
```
The naming is clear, but given this is somewhat complex async behavior, I would add these two lines. I'm not sure waitAll is useful tbh, as it only is useful if the spinner terminates on its own AND does never throw, which seems a non-existing case. While it is elegant and adds completeness, I feel like removing it.

lhruby (approved):

modules/concurrency/include/hephaestus/concurrency/spinner_manager.h
filippobrizzi: you are right, it could be indeed an early generalization having `waitAll`. 
The current problem with this interface is that from outside (until we call `.get()`) we have no way to know if something was terminated with an exception. In addition, it is not given that just because one of the spinners terminated with an exception you want to stop all of them.
For this reason, you also want to assume there are cases where you want to wait for all the spinner to complete. 
Given the simplicity of the implementation and the fact that it doesn't have impact to the overall design I would keep `waitAll`. WDYT?

lhruby: sounds good

lhruby (approved):

PR #200, opened 2024-11-08
Feat/Added protobuf serialization for BitField

lievenvanderheide:
# Description
- Moved BitFlag to the containers module.
- Added a constructor to BitFlag which takes the underlying value.
- Added a containers_proto mudole
- Implemented protobuf serialization of BitFlag

## Type of change
- Breaking change - BitFlag is now in a different namespace, so existing code may not compile.


modules/containers/include/hephaestus/containers/bit_flag.h
```diff
@@ -66,6 +82,22 @@ class BitFlag {
                   "Enum is not valid for BitFlag, its values must be power of 2.");
   }
 
+  /// Constructs a BitFlag with the given underlying value.
+  ///
+  /// An 'InvalidParameterException' is thrown if there are bits set in 'underlying_value' which do not
+  /// correspond to a valid enum value.
+  constexpr explicit BitFlag(std::underlying_type_t<EnumT> underlying_value) : value_(underlying_value) {
```
filippobrizzi, line 89: nit. we define `T` as the underlying type so we could use it here. If you think the name is not expressive enough, feel free to refactor to a better one

modules/containers_proto/proto/hephaestus/containers/proto/bit_flag.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package heph.containers.proto;
+
+message BitFlag {
+  uint64 value = 1;
+}
```
filippobrizzi, line 7: missing empty line at the end of the file. You can add a setting to your VS Code to automatically add an empty line

modules/containers_proto/include/hephaestus/containers_proto/bit_flag.h
```diff
@@ -0,0 +1,31 @@
...
+struct ProtoAssociation<containers::BitFlag<EnumT>> {
+  using Type = containers::proto::BitFlag;
+};
+}  // namespace heph::serdes::protobuf
+
+namespace heph::containers {
+
+template <typename EnumT>
+auto toProto(proto::BitFlag& proto_bit_flag, const BitFlag<EnumT>& bit_flag) -> void {
+  proto_bit_flag.set_value(bit_flag.getUnderlyingValue());
```
filippobrizzi, line 23: Don't we need a static_Cast here as well?

modules/containers_proto/include/hephaestus/containers_proto/bit_flag.h
```diff
@@ -0,0 +1,31 @@
...
+}
+
+template <typename EnumT>
+auto fromProto(const proto::BitFlag& proto_bit_flag, BitFlag<EnumT>& bit_flag) -> void {
+  bit_flag = BitFlag<EnumT>{ static_cast<std::underlying_type_t<EnumT>>(proto_bit_flag.value()) };
```
filippobrizzi, line 28: nit.
```suggestion
  bit_flag = BitFlag<EnumT>{ static_cast<BitFlag<EnumT>::T>(proto_bit_flag.value()) };
```

modules/containers_proto/tests/bitflag_serialization_tests.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+// NOLINTNEXTLINE(google-build-using-namespace)
+using namespace ::testing;
+
+namespace heph::containers::tests {
+
+TEST(BitFlagSerialization, RoundTrip) {
+  enum class Enum : uint8_t { A = 1u << 3u, B = 1u << 5u };
+
+  const BitFlag<Enum> value{ Enum::B };
```
filippobrizzi, line 21: probably unnecessary but let's have a composed value here

modules/containers_proto/include/hephaestus/containers_proto/bit_flag.h
lievenvanderheide: It's not necessary, because casting from an int to an int of larger size is implicit. I could still add it, to make it clear a conversion is happening if you want.

filippobrizzi: ah good point, also if clang-tidy doesn't complain I think we are good

filippobrizzi (approved):

PR #203, opened 2024-11-13
[Fix] pub/sub and service now check for underlying serialized type instead of the top level one

filippobrizzi:
# Description
When subscribing or calling a service we check that we are getting the expected underlying type.
This change is needed when bridging with ROS as we actually deserialize the data in a different type, and also this is the correct logic.


lhruby (approved):
LGTM, thx.

PR #205, opened 2024-11-14
Fix/Add wait to message consumer and example

johannes-graeter:
# Description
Add a wait function for message consumer.
This is important to have the metric and also the logging wait until the message queue is empty before destructing and hence stopping the consumer.
When commenting out the wait you can check this by calling the provided metric example. You might need to execute it several times in order to observe that sometimes the metric is printed sometimes not.

When adding the wait it should be printed always.

## Type of change

- [x ] I have performed a self-review of my code.
- [x] If it is a core feature, I have added thorough tests. -> already existed
- [x ] If this is a new component I have added examples.


johannes-graeter:
I assume when deactivating exceptions, the destructor would directly return with a fatal error if anything goes wrong in wait() right?

modules/concurrency/include/hephaestus/concurrency/message_queue_consumer.h
```diff
@@ -58,6 +59,13 @@ void MessageQueueConsumer<T>::stop() {
   spinner_.stop();
 }
 
+template <typename T>
+void MessageQueueConsumer<T>::wait() {
+  while (!message_queue_.empty()) {
+    consume();
+  }
+}
```
filippobrizzi, line 67: I think this logic should be moved inside the `stop` function:
```
template <typename T>
void MessageQueueConsumer<T>::stop() {
  message_queue_.stop(); // We do not accept messages anymore
  spinner_.stop(); // We make sure that the spinner is no longer accessing the queue
 // We go through all the messages left and we consume them
  while (!message_queue_.empty()) {
    consume();
  }
}
```
At this point to make it perfect I would also probably make `stop` return a `std::future` so that we do not block while stopping.

Also I would remove the `wait` queue all together from the public interface as it can break the spinner or the ordering.

filippobrizzi (commented):
Thanks! Left a comment

modules/concurrency/include/hephaestus/concurrency/message_queue_consumer.h
```diff
@@ -25,7 +27,10 @@ class MessageQueueConsumer {
   auto operator=(MessageQueueConsumer&&) -> MessageQueueConsumer& = delete;
 
   void start();
-  void stop();
+
+  /// @brief Stop the message consumer, emptying the queue and stopping the processing.
+  /// @return future that waits on the queue to be emptied
+  auto stop() -> std::future<void>;
```
filippobrizzi, line 33:
```suggestion
  [[nodiscard]] auto stop() -> std::future<void>;
```

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -126,7 +127,11 @@ MetricRecorder::MetricRecorder()
 }
 
 MetricRecorder::~MetricRecorder() {
-  entries_consumer_.stop();
+  try {
+    entries_consumer_.stop().wait();
```
filippobrizzi, line 131:
```suggestion
    entries_consumer_.stop().get();
```

johannes-graeter: whats the difference in this case? return value is void right?

modules/concurrency/include/hephaestus/concurrency/message_queue_consumer.h
johannes-graeter: great, much better design!

modules/telemetry/telemetry/src/metric_record.cpp
filippobrizzi: you need to call `get` on a future to surface potential exception from inside the asynch

johannes-graeter: crazy so exceptions are suppressed when calling wait()?

johannes-graeter:
add ing [nodiscard] to stop changes behavour in many places, but to me it seems this is a fix rather than a bug right?

modules/telemetry/telemetry/src/metric_record.cpp
filippobrizzi: not really, the point of `wait` is that eventually you call `get`. You can call `wait` many time on the same object, but once you call `get` the future is no longer valid.

filippobrizzi (approved):
nice!

PR #206, opened 2024-11-14
Feat/Add a state machine to Spinner.

lhruby:
# Description
Add a state machine to `Spinner` to implement lifecycle management. 

Lifecycle management is the structured process of managing the states and transitions of components or applications throughout their existence to ensure efficient resource allocation and robust operation. A popular application in lifecycle management is a component’s recovery from a failure case without termination.

state_machine drawio

## Type of change


modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -11,24 +11,37 @@
 
 namespace heph::concurrency {
 
+struct SpinnerCallbacks {
+  std::function<void() init_cb = []() {};
+  std::function<void() spin_once_cb = []() {};
```
filippobrizzi, line 16: what about the StoppableCallback one? We do not want to support in this case? 
also you are missing:
```suggestion
  std::function<void()> spin_once_cb = []() {};
```

modules/concurrency/src/spinner.cpp
```diff
@@ -64,34 +56,76 @@ void Spinner::spin() {
 
   try {
     start_timestamp_ = std::chrono::system_clock::now();
+    uint64_t spin_count_{ 0 };
 
     while (!stop_requested_.load()) {
-      const auto spin_result = stoppable_callback_();
+      // State machine to handle the spinner lifecycle.
```
filippobrizzi, line 62: here the code is a bit too convoluted and hard to follow, I would try to first break it down into steps:
(I do not even think you need the states in the enum).
```c++
void Spinner::spin() {
  // This is the main loop of the whole state machine
  while (!stop_requested_)) {
    auto res = initialize();
    if (!res) { break; } // This means we have trying initializing for a while and we couldn't so there is nothing more we can do
    auto res = execute();
    if (shall_stop_cb(res)) break; // We restart from init
    else continue;
  }
  termination_cb();
}

auto Spinner::initialize() -> bool {
  while (!stop_requested_) {
     try {
       init_cb()
       return true;
     } catch() {
        if (should_re_init()) continue;
        else return false;
     }
  }
}

auto Spinner::execute() -> bool {
  try {
     while (!stop_requested_)) {
        spin_cb()
         std::this_thread::sleep_for(...);
     }
  } catch() {
     return false;
  }
}

```

lhruby: This is rather a conceptual question. Your solution removes observability from the states, and it has some implications
* all transitions are always executed, unless we add additional checks
* how would you implement these checks? e.g. is_initialized? I assume you would add a state
* the proposed solution works as long as the logic chain is linear, but does not allow for branching, see. e.g. the missing branching surrounding `stop?`. I see that we can still fix it for the current logic, but extendability in the future will suffer.
* Extendability: we extended `Spinner` to `StoppableSpinner`. Now we extend to `ReInitableAndStoppableSpinner`. There's a good chance we will add additional capabilities in the future. I thus prefer an extendable design.

modules/concurrency/include/hephaestus/concurrency/spinner.h
lhruby: fixed

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -11,24 +11,37 @@
...
-  enum class SpinResult : bool { CONTINUE, STOP };
-  using StoppableCallback = std::function<SpinResult()>;
-  using Callback = std::function<void()>;
-
-  /// @brief Create a spinner with a stoppable callback. A stoppable callback is a function that returns
-  /// SpinResult::STOP to indicate that the spinner should stop.
-  /// Example: a callback that stops after 10 iterations.
-  explicit Spinner(StoppableCallback&& stoppable_callback, double rate_hz = 0);
-
+  enum class SpinnerState : uint8_t {
```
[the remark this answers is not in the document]
lhruby: renamed

modules/concurrency/src/spinner.cpp
filippobrizzi: If you are in `execute` you can assume that you have successfully initialized.
I think what confuses me is that in your solution you have a single loop that handles all the states. I am not a state machine expert so maybe that is the correct way to go, but it make also the code much more complex. 
I was proposing a more hierarchical solution. Where you have a top level loop and then inside you run the different states.
I am not sure I understand the missing branch surrounding stop. In my proposal you check if you want to start from scratch or not. 

In my opinion, we either take the time and implement a proper nice state machine abstraction, maybe even check at some 3rd party for reference, or it is fine to avoid early generalization and overfit on our problem at end.

modules/concurrency/src/spinner.cpp
```diff
@@ -64,34 +81,60 @@ void Spinner::spin() {
...
+
+      state_ = attemptBinaryTransition(state_, { .input_state = State::READY_TO_SPIN,
+                                                 .operation = callbacks_.spin_once_cb,
+                                                 .success_state = State::SPIN_SUCCESSFUL,
+                                                 .failure_state = State::SPIN_FAILED });
+
+      state_ = attemptBinaryTransition(state_, { .input_state = State::SPIN_FAILED,
+                                                 .operation = callbacks_.shall_re_init_cb,
+                                                 .success_state = State::NOT_INITIALIZED,
+                                                 .failure_state = State::TERMINATE });
```
filippobrizzi, line 105: I think `shall_re_init_cb` should be associated only to init failures, here we should call, `shall_stop`

modules/concurrency/src/spinner.cpp
```diff
@@ -64,34 +81,60 @@ void Spinner::spin() {
...
         break;
       }
 
+      // Logic to handle periodic spinning at a fixed rate.
       if (spin_period_.count() == 0) {
         continue;
       }
-
-      const auto target_timestamp = start_timestamp_ + spin_count_ * spin_period_;
+      const auto target_timestamp = start_timestamp_ + (++spin_count_ * spin_period_);
```
filippobrizzi, line 120: There is an issue here. This logic doesn't old true anymore as we also execute init code inside here. We need to update to keep track only of the time spent spinning.

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -40,27 +43,29 @@ class Spinner {
   auto stop() -> std::future<void>;
   void wait();
 
-  /// @brief  Set a callback that will be called when the spinner is stopped.
-  /// This callback could be extendend to pass the reason why the spinner was stopped, e.g. exceptions, ...
-  void setTerminationCallback(Callback&& termination_callback);
```
filippobrizzi, line 45: This function is used by the SpinnerManager. If we remove it we cannot use the Spinner manager. this is because the spinner manager receive a spinner and need to be able to inject a termination callback,

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -17,18 +17,21 @@ namespace heph::concurrency {
...
+  using TransitionCallback = std::function<void()>;
+  using PolicyCallback = std::function<bool()>;
 
-  /// @brief Create a spinner with a stoppable callback. A stoppable callback is a function that returns
-  /// SpinResult::STOP to indicate that the spinner should stop.
-  /// Example: a callback that stops after 10 iterations.
-  explicit Spinner(StoppableCallback&& stoppable_callback, double rate_hz = 0);
+  struct Callbacks {
+    TransitionCallback init_cb = []() {};
+    TransitionCallback spin_once_cb = []() {};
```
filippobrizzi, line 25: I think we can still support the stoppable callback with this design

lhruby: We do, we just change the approach. Instead of injecting it into `spin()` via the executable callback, it's now a first class citizen via `shall_stop_cb`.

modules/concurrency/src/spinner.cpp
lhruby: Can I close this one @filippobrizzi ?

(The proposed code is very elegant by the way, just wanted to point that out)

modules/concurrency/include/hephaestus/concurrency/spinner_state_machine.h
```diff
@@ -0,0 +1,30 @@
+//=================================================================================================
+// Copyright (C) 2023-2024 HEPHAESTUS Contributors
+//=================================================================================================
+
+#pragma once
```
filippobrizzi, line 5: let's keep this in the same file, it is much easier to read without having to jump between files together they are less than 100 loc

modules/concurrency/src/spinner_state_machine.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+namspace {
+  template <typename Callback>
+  struct BinaryTransitionParams {
+    State input_state;
+    Callback& operation;
+    State success_state;
+    State failure_state;
+  };
+  [[nodiscard]] auto attemptBinaryTransition(State current_state, const BinaryTransitionParams& params)
```
filippobrizzi, line 17:
```suggestion
  };
  
  [[nodiscard]] auto attemptBinaryTransition(State current_state, const BinaryTransitionParams& params)
```

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -13,22 +13,37 @@ namespace heph::concurrency {
...
 public:
   enum class SpinResult : bool { CONTINUE, STOP };
   using StoppableCallback = std::function<SpinResult()>;
   using Callback = std::function<void()>;
 
-  /// @brief Create a spinner with a stoppable callback. A stoppable callback is a function that returns
-  /// SpinResult::STOP to indicate that the spinner should stop.
-  /// Example: a callback that stops after 10 iterations.
-  explicit Spinner(StoppableCallback&& stoppable_callback, double rate_hz = 0);
+  /// @brief Create a callback for the spinner which never stops.
```
filippobrizzi, line 24:
```suggestion
  /// @brief Wrap the user provided callback in a callback that never stops.
```

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -13,22 +13,37 @@ namespace heph::concurrency {
...
+  [[nodiscard]] static auto createNeverStoppingCallback(Callback&& callback) -> StoppableCallback;
+
+  struct StateMachineCallbacks {
+    using TransitionCallback = std::function<void()>;
+    using PolicyCallback = std::function<bool()>;
+
+    TransitionCallback init_cb = []() {};       //!< Handles initialization.
+    TransitionCallback spin_once_cb = []() {};  //!< Handles execution.
```
filippobrizzi, line 32: should we call this? It is more general meaning, but it is also true that we are in a spinner, so probably spin_once is more correct/clear
```suggestion
    TransitionCallback execute_cb = []() {};  //!< Handles execution.
```

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -13,22 +13,37 @@ namespace heph::concurrency {
...
+
+    PolicyCallback shall_stop_spinning_cb = []() { return false; };  //!< Default: spin indefinitely.
```
filippobrizzi, line 34: Let's add a comment that this callback is called after a successful execution.

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -13,22 +13,37 @@ namespace heph::concurrency {
...
+    PolicyCallback shall_restart_cb = []() { return false; };        //!< Default: do not restart.
```
filippobrizzi, line 35: Let's add a comment that this is called after failure

modules/concurrency/include/hephaestus/concurrency/spinner.h
```diff
@@ -13,22 +13,37 @@ namespace heph::concurrency {
...
-  /// Example: a callback which reads data from a sensor, until the spinner is stopped.
-  explicit Spinner(Callback&& callback, double rate_hz = 0);
+  /// @brief Create a callback for the spinner which internally handles the state machine.
+  [[nodiscard]] static auto createCallbackWithStateMachine(StateMachineCallbacks&& callbacks)
+      -> StoppableCallback;
+
+  /// @brief Create a spinner with a stoppable callback. A stoppable callback is a function that returns
+  /// SpinResult::STOP to indicate that the spinner should stop. Other types of callbacks are supported via
+  /// mappings to StoppableCallback. Example: a callback that stops after 10 iterations.
+  explicit Spinner(StoppableCallback&& stoppable_callback,
```
filippobrizzi, line 45: Question: didn't we decide that we actually do not need the stoppable callback as we can use `shall_stop_spinning_cb` for this case? In case we use the stoppable callback, shouldn't we remove `shall_stop_spinning_cb`?

lhruby: The stoppable callback is needed to separate the state machine from the spinner.

The spinner spins forever (`Spinner::SpinResult::Continue`), until it is told by the callback to exit (`Spinner::SpinResult::Stop`). We don't impose a reason for the stop, which can either be a simple StoppableCallback where the iteration counter has elapsed, or it can be a callback stemming from a state machine, where the state machine reached exit.

modules/concurrency/include/hephaestus/concurrency/spinner.h
lhruby: done

filippobrizzi (approved):

PR #207, opened 2024-11-19
Feat/multicast scouting settings

Fellfalla:
# Description
Currently, default settings for multicast scouting is used which is enabled and automatically selects network devices for scouting. This could lead to issues on unstable networks. The new config parameters can be used to disable multicast scouting entirely or to define a specific multicast interface.

## Type of change
Configuration feature


Fellfalla:
I wonder why we don't use `::zenoh::Config::from_file` to parse settings from file, instead of using default settings?

filippobrizzi (approved):

PR #211, opened 2024-11-22

johannes-graeter:
# Description

In succession of #201 this is my first try to implement the log interface like
```cpp
heph::log(heph::Level::WARN, "message", {"num", 123}, {"test", "lala"});
```
I tried doing the variadic interface but struggled to get the template deduction guides working that are needed to combine the variadic interface with source_location  with direct intitialization via {}.

Any ideas?

another way would be to have a class
```cpp
struct MessageWithLocation {
  std::string value;
  std::source_location loc;

  MessageWithLocation(std::string s,
                     std::source_location l = std::source_location::current())
      : value(s), loc(l) {}

};
```
that would be used instead of std::string for message input type. However I am not sure if then implicit conversion from string would be possible (`heph::log(heph::Level::WARN, "message")` vs `heph::log(heph::Level::WARN, {"message"})`
## Type of change

### DIFF
--- a/modules/telemetry/telemetry/CMakeLists.txt
+++ b/modules/telemetry/telemetry/CMakeLists.txt
@@ -11,14 +11,14 @@ declare_module(
 find_package(fmt REQUIRED)
 
 # library sources
-set(SOURCES src/metric_sink.cpp src/metric_record.cpp src/metric_sinks/terminal_sink.cpp include/hephaestus/telemetry/metric_sink.h
-            include/hephaestus/telemetry/metric_record.h include/hephaestus/telemetry/metric_sinks/terminal_sink.h
+set(SOURCES src/metric_sink.cpp src/metric_record.cpp src/struclog.cpp src/log_sinks/absl_sink.cpp src/metric_sinks/terminal_sink.cpp 
+            include/hephaestus/telemetry/metric_sink.h include/hephaestus/telemetry/metric_record.h include/hephaestus/telemetry/struclog.h include/hephaestus/telemetry/log_sinks/absl_sink.h include/hephaestus/telemetry/metric_sinks/terminal_sink.h
 )
 
 # library target
 define_module_library(
   NAME telemetry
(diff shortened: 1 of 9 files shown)


filippobrizzi:
@johannes-graeter thanks for the effort! what seemed like a normal task is becoming a fight for humanity.

I am starting to think that there is no escape from macros. 

 `ht::log(ht::Level::INFO, "testing absl log sink with field", FIELD(num));`

I tried briefly to make it work with `MessageWithLocation` but I didn't have success. I can try more in the next days.

@lievenvanderheide do you happen to have any idea how we could make this work?

johannes-graeter:
There is a few more things I would like to try.
A. In the oroginal article I got this from https://stackoverflow.com/questions/57547273/how-to-use-source-location-in-a-variadic-template-function#57548488 the message with location is not a string but a const char* and if the example is valid it is directly initializable... perhaps this is a straw
B. Interface without {} so
```cpp
log(Level::WARN, "message", "field1", field1, "field2", field2)
```
Don't really like it since I am missing visual separation of the fields but better than using macros...

filippobrizzi:
> There is a few more things I would like to try. A. In the oroginal article I got this from https://stackoverflow.com/questions/57547273/how-to-use-source-location-in-a-variadic-template-function#57548488 the message with location is not a string but a const char* and if the example is valid it is directly initializable... perhaps this is a straw B. Interface without {} so
> 
> ```c++
> log(Level::WARN, "message", "field1", field1, "field2", field2)
> ```
> 
> Don't really like it since I am missing visual separation of the fields but better than using macros...

You are right this (`log(Level::WARN, "message", "field1", field1, "field2", field2)`) would look better than a macro, although it could cause issues if user forget to pass the variable.

filippobrizzi:
Amazing work @johannes-graeter, I think I like `PropA` the best, it is slim and clear. There is still the risk of missing a value and mixing keys with values, but that I think is an acceptable risk.

PR #213, opened 2024-11-25

johannes-graeter:
# Description
Implemented the interface discussed in #211 and #201.
useable as 
```cpp
  const int num = 1234;
  heph::telemetry::log(heph::telemetry::Level::WARN, "testing absl log sink");
  heph::telemetry::log(heph::telemetry::Level::WARN, "testing absl log sink with field", "num", num, "quoted_string", "test");
```
see tests and examples.

## Type of change


johannes-graeter:
@lhruby @filippobrizzi @floriantschopp one remaining issue is if we should elevate `heph::telemetry::log` and Level to `heph::log`?

johannes-graeter:
For uneven parameters I added 'static_assert(false,"")' to enhance readability of the compile error in case the parameter are mixed up by the user. However the build_gcc staged failed then, but in the example in the comment above it worked fine with gcc 14.2 and cpp20.
Making this work without dirty hacks is newer according to https://stackoverflow.com/questions/76673888/why-is-always-false-v-required-in-this-situation I am AFK but I guess the gcc version in ci could be updated?

modules/telemetry/telemetry/include/hephaestus/telemetry/struclog.h
```diff
@@ -0,0 +1,187 @@
...
+    // Pointer decay is wanted here to catch char[n] messages
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-array-to-pointer-decay)
+    ss << std::quoted(field.value);
+    fields.emplace_back(field.key, ss.str());
+
+    return std::move(*this);
+  }
+
+  Level level;
+  MessageWithLocation msg_with_loc;
```
filippobrizzi, line 102: Can we unpack message and location here?

modules/telemetry/telemetry/include/hephaestus/telemetry/struclog.h
```diff
@@ -0,0 +1,187 @@
...
+struct ILogSink {
+  using Formatter = std::function<std::string(const LogEntry& log)>;
+
+  virtual ~ILogSink() = default;
+
+  /// @brief Main interface to the sink, gets called on log.
+  ///
+  /// @param log_entry
+  virtual void send(const LogEntry& log_entry) = 0;
+};
```
filippobrizzi, line 132: please move this into `log_sink.h`

modules/telemetry/telemetry/src/log_sinks/absl_sink.cpp
```diff
@@ -0,0 +1,44 @@
...
+      break;
+    case Level::ERROR:
+      ABSL_LOG(ERROR).NoPrefix() << formatter_(entry);
+      break;
+    case Level::FATAL:
+      ABSL_LOG(FATAL).NoPrefix() << formatter_(entry);
+      break;
+  }
+}
+}  // namespace heph::telemetry
```
filippobrizzi, line 44: missing end-of-line

modules/telemetry/telemetry/include/hephaestus/telemetry/struclog.h
```diff
@@ -0,0 +1,187 @@
...
+}
+
+///@brief Log a message without fields. Example:
+///       ```
+///       log(Level::WARN, "speed is over limit", "current_speed"))
+///       ```
+template <typename... Args>
+void log(Level level, MessageWithLocation msg) {
+  internal::log(LogEntry{ level, msg });
+}
```
filippobrizzi, line 185: let's place this in `namespace heph`

filippobrizzi:
> For uneven parameters I added 'static_assert(false,"")' to enhance readability of the compile error in case the parameter are mixed up by the user. However the build_gcc staged failed then, but in the example in the comment above it worked fine with gcc 14.2 and cpp20. Making this work without dirty hacks is newer according to https://stackoverflow.com/questions/76673888/why-is-always-false-v-required-in-this-situation I am AFK but I guess the gcc version in ci could be updated?

We could add 
```c++
#ifndef __GNUC__
// TODO: remove ifdef once we migrate to GCC 14
...
#endf
```

filippobrizzi (commented):
Fantastic work!! please check my comments

johannes-graeter:
I also elevated the Log level as to write less to `heph::Error`is that too ambiguitive? perhaps we can have a short namespace in between like heph::loglevel::Error or use a name like `heph::ErrorLevel`or `heph::ErrorL`
@filippobrizzi WDYT?

filippobrizzi:
> I also elevated the Log level as to write less to `heph::Error`is that too ambiguitive? perhaps we can have a short namespace in between like heph::loglevel::Error or use a name like `heph::ErrorLevel`or `heph::ErrorL` @filippobrizzi WDYT?

What about `heph::LogLevel::INFO`?

modules/telemetry/telemetry/include/hephaestus/telemetry/struclog.h
johannes-graeter: done for this and LogEntry (same as metric_sink and metric_record) also renamed struclog.h to log.h

johannes-graeter:
> > For uneven parameters I added 'static_assert(false,"")' to enhance readability of the compile error in case the parameter are mixed up by the user. However the build_gcc staged failed then, but in the example in the comment above it worked fine with gcc 14.2 and cpp20. Making this work without dirty hacks is newer according to https://stackoverflow.com/questions/76673888/why-is-always-false-v-required-in-this-situation I am AFK but I guess the gcc version in ci could be updated?
> 
> We could add
> 
> ```c++
> #ifndef __GNUC__
> // TODO: remove ifdef once we migrate to GCC 14
> ...
> #endf
> ```

I have added `#if (__GNUC__ >= 14) || defined(__clang__)`

johannes-graeter:
One thing that makes me still a little uneasy is that now the message is not owned by LogEntry anymore. For literals this doesn't not matter as stated above since they are static, but it would be easy to get UB if you would decide to compose the message before as string and give over the c_str... this is not the usual usage, but could still happen.

It would be safer to store the message as std::string, but I am uncertain if the compiler would inline the static string literal in normal usage or if it would copy...

Do you know that @filippobrizzi or @lievenvanderheide ?

johannes-graeter:
> > One thing that makes me still a little uneasy is that now the message is not owned by LogEntry anymore. For literals this doesn't not matter as stated above since they are static, but it would be easy to get UB if you would decide to compose the message before as string and give over the c_str... this is not the usual usage, but could still happen.
> 
> > 
> 
> > It would be safer to store the message as std::string, but I am uncertain if the compiler would inline the static string literal in normal usage or if it would copy. For large strings I am pretty certain that it will copy but for short ones, I don't know...
> 
> > 
> 
> > Do you know that @filippobrizzi or @lievenvanderheide ?
> 
> 
> 
> The reason why `MessageWithLocation` doesn't save a `std::string` is to avoid copies? 

So the constructor of MessageWithLocation needs a const char* as input otherwise "my message" will not implicitly casted to it. 
It does not store as string to avoid copies yes.
I guess if we use "my message"s we could have a string in the constructor... we could the message type everywhere and enforce << through a concept but I am unsure if that is not overengeneered...

modules/telemetry/telemetry/examples/log_sink_example.cpp
```diff
@@ -0,0 +1,23 @@
...
+// NOLINTNEXTLINE(bugprone-exception-escape)
+auto main(int /*argc*/, const char* /*argv*/[]) -> int {
+  const heph::utils::StackTrace stack;
+
+  heph::telemetry::registerLogSink(std::make_unique<heph::telemetry::AbslLogSink>());
+
+  heph::log(heph::LogLevel::WARN, "testing absl log sink");
+
+  const int num = 1234;
+  heph::log(heph::LogLevel::INFO, "absl log sink with fields", "num", num, "quoted_string", "test");
```
floriantschopp, line 20: great, thanks guys for adjusting, how about the fields being pairs, ie.e 
```suggestion
  heph::log(heph::LogLevel::INFO, "absl log sink with fields", {"num", num}, {"quoted_string", "test"});
```
makes it clearer to what belongs together in a key-value way, no?

floriantschopp (approved):

modules/telemetry/telemetry/examples/log_sink_example.cpp
johannes-graeter: yes definitely, this was our prefered solution, however I struggled combining brace initializers with variadics...
see https://github.com/olympus-robotics/hephaestus/pull/211
what your take, would you prefer
```
heph::log(heph::LogLevel::INFO, "absl log sink with fields", {{"num", num}, {"quoted_string", "test"}});
```
? @filippobrizzi and I prefered the interface with less to write...

johannes-graeter:
@filippobrizzi I added an additional constructor supporting std::string to MessageWithLocation.
This has 2 advantages:
A. using string literals ""s we can use the interface without copying in case we have long strings
B. we can use log(LogLevel::Warn, std::format("{}",123)); without copies (see https://github.com/olympus-robotics/hephaestus/pull/201#discussion_r1834707786)

johannes-graeter:
since commit [2bad994](https://github.com/olympus-robotics/hephaestus/pull/213/commits/2bad994d308c248d98bb0a206373c7550bb2b786) I am struggling with the fact that fmt/core.h seems deprecated now. Looks as if https://github.com/olympus-robotics/hephaestus/pull/214 solved this. 
If some fix gets merged before this, we should reset to [8479ea0](https://github.com/olympus-robotics/hephaestus/pull/213/commits/8479ea0ed2009b7136e4347c2ee856caa6ea6ed8) and then merge main.

modules/telemetry/telemetry/examples/log_sink_example.cpp
floriantschopp: thats ok, lets keep the above

johannes-graeter:
blocked by https://github.com/olympus-robotics/hephaestus/pull/214

modules/telemetry/telemetry/include/hephaestus/telemetry/log_sink.h
```diff
@@ -0,0 +1,145 @@
...
+#include <source_location>
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <utility>
+#include <vector>
+
+namespace heph {
+enum class LogLevel : std::uint8_t { TRACE, DEBUG, INFO, WARN, ERROR, FATAL };
```
filippobrizzi, line 18: Am I crazy if I suggest:
```suggestion
enum LogLevel : std::uint8_t { TRACE, DEBUG, INFO, WARN, ERROR, FATAL };
```
so that we can do: `heph::log(heph::INFO, ...)` 

This symbols are very log specific, I doubt we will cause any clashes

modules/telemetry/telemetry/src/log_sink.cpp
```diff
@@ -0,0 +1,77 @@
...
+                                log.location.line()));
+  ss << " thread-id=" << log.thread_id;
+  ss << " time=" << fmt::format("{:%Y-%m-%dT%H:%M:%SZ}", log.time);
+  ss << " message=" << std::quoted(log.message);
+
+  for (const Field<std::string>& field : log.fields) {
+    ss << " " << field.key << "=" << field.value;
+  }
+
+  return ss.str();
```
filippobrizzi, line 74: nit. I am wondering if it is not better to use `fmt::format`, I know you cannot use `std::quoted` but it should be still easy to add them with `\"\"` and it is more inline with the rest of the formatting in hephaestus.

Also for field you can use `join`.

modules/telemetry/telemetry/include/hephaestus/telemetry/log_sink.h
johannes-graeter: I actually also considered it, however I am realy not a fan of implicit cast of enums to uchar...
we could consider strong typing, just defining them as individual structs and enforcing them with a concept like
https://godbolt.org/z/nG1G4PrGb
might be overkill though

modules/telemetry/telemetry/src/log_sink.cpp
johannes-graeter: adding  ` \"\"` is unfortuanetly not the same behaviour, since std::quoted would not quote the string if it is already quoted (to get rid of double quotes). So we would need to check if the user quoted the quotable which is also easy enough but would add some boiler plate. Also we can just life with the fact that the user may not quote the message or gets double quotes...

modules/telemetry/telemetry/include/hephaestus/telemetry/log.h
```diff
@@ -0,0 +1,75 @@
...
+#include <memory>
+#include <sstream>
+#include <string>
+#include <utility>
+
+#include "hephaestus/telemetry/log_sink.h"
+
+namespace heph::telemetry {
+
+void registerLogSink(std::unique_ptr<ILogSink> sink);
```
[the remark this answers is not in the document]
johannes-graeter: so in the heph namespace right?

modules/telemetry/telemetry/src/log_sink.cpp
johannes-graeter: i now added fmt for everything, also replaced the format() function by a fmt::formatter to be consistent (using fmt as default for formatting everywhere...).
only downside is that the formatting now all resides in the header... what do you think? @filippobrizzi

modules/telemetry/telemetry/include/hephaestus/telemetry/log_sink.h
filippobrizzi: This is a cool concept, but I think we can live with having a non class enum. It is for a very specific case, so it should be ok.

modules/telemetry/telemetry/include/hephaestus/telemetry/log.h
```diff
@@ -0,0 +1,73 @@
...
+///       ```
+///       heph::log(heph::LogLevel::WARN, "speed is over limit")
+///       ```
+template <typename... Args>
+void log(LogLevel level, telemetry::MessageWithLocation&& msg) {
+  telemetry::internal::log(telemetry::LogEntry{ level, std::move(msg) });
+}
+
+///@brief Register a sink for logging.
+void registerLogSink(std::unique_ptr<telemetry::ILogSink> sink);
```
filippobrizzi, line 71: sorry if it wasn't clean I would keep this here, but maybe I would still wrap it in `heph::telemetry` namespace.

modules/telemetry/telemetry/include/hephaestus/telemetry/log_sink.h
```diff
@@ -0,0 +1,171 @@
...
+      case heph::LogLevel::WARN:
+        return fmt::format_to(ctx.out(), "warn");
+      case heph::LogLevel::ERROR:
+        return fmt::format_to(ctx.out(), "error");
+      case heph::LogLevel::FATAL:
+        return fmt::format_to(ctx.out(), "fatal");
+      default:
+        return fmt::format_to(ctx.out(), "XXX");
+    }
+  }
```
filippobrizzi, line 158: why not using `magic_enum`?

filippobrizzi (commented):
I think we are ready! last 2 comments and then we can merge!

modules/telemetry/telemetry/include/hephaestus/telemetry/log_sink.h
johannes-graeter: oh yeah awesome!

johannes-graeter:
Added suggestions and ported to bazel. Took the liberty to add the metric_sink_example as binary.

modules/telemetry/telemetry/BUILD
```diff
@@ -35,3 +36,31 @@ heph_cc_test(
...
+
+heph_cc_test(
+    name = "log_tests",
+    srcs = ["tests/log_tests.cpp"],
+    deps = [
+        ":telemetry",
+    ],
+)
+
+############
```
filippobrizzi, line 51: Very nit. but I wouild put binaries before tests, so that it's easier to see them

filippobrizzi (approved):
Fantastic work!

johannes-graeter:
started adding the logs in branch feature/port-to-hephlog

PR #217, opened 2024-11-29
Spinner: fix period logic

filippobrizzi:
# Description
Fix logic for the periodic spinner so that if a call to `spinOnce` takes longer than the period it doesn't mess with the following  ticks. 

The change is that we do not long keep track of spin count, but we compute it live based on the elapsed time and the period


lhruby:
Nice

lhruby (approved):

PR #220, opened 2024-12-05
Fix/Fix nondetermenistic test for logging

johannes-graeter:
# Description

With using several sink, the asynchronous global sink became non deterministic, since it only held the last log.
Now it supports having several logs. and blocks until they are available. 
Tried to do it elegantly, with condition variable, but failed and did it with a sleeping loop.
@filippobrizzi can you get the condition variable to work? I always ended up in a deadlock...

## Type of change

### DIFF
--- a/modules/telemetry/telemetry/tests/log_tests.cpp
+++ b/modules/telemetry/telemetry/tests/log_tests.cpp
@@ -1,10 +1,14 @@
 //=================================================================================================
 // Copyright (C) 2023-2024 HEPHAESTUS Contributors
-#include <atomic>
+
+#include <chrono>
+// #include <condition_variable>
 #include <format>
 #include <iomanip>
 #include <iostream>
 #include <memory>
+// #include <mutex>
+#include <source_location>
 #include <sstream>
 #include <string>
 #include <utility>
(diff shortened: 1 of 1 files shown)


modules/telemetry/telemetry/tests/log_tests.cpp
```diff
@@ -18,11 +22,50 @@
...
+*/
+    using namespace std::chrono_literals;
+    while (logs_.find(loc.line()) == logs_.cend()) {
+      std::this_thread::sleep_for(10ms);
+    }
+    return logs_.at(loc.line());
+  }
+
+private:
+  std::map<Line, std::string> logs_;
```
filippobrizzi, line 52: You can use a blocking queue. From the test it looks like you call `log` and then `getLog` so this should work. It will block until you have a log message available.
```suggestion
  void send(const ht::LogEntry& s) override {
    logs_.emplaceBack(fmt::format("{}", s));
  }

  [[nodiscard]] auto getLog() -> std::string {
      return logs_.waitAndPop().value();
  }

private:
    containers::BlockingQueue<std::string> logs_;
```

modules/telemetry/telemetry/tests/log_tests.cpp
```diff
@@ -72,118 +114,79 @@ TEST(log, Escapes) {
...
+  // clang-format on
+
   ht::internal::log(std::move(log_entry));
-  sink_ptr->wait();
+  fmt::print("{}", sink_ptr->getLog(loc));
   {
     std::stringstream ss;
     ss << "num=" << num;
-    EXPECT_TRUE(sink_ptr->getLog().find(ss.str()) != std::string::npos);
+    EXPECT_TRUE(sink_ptr->getLog(loc).find(ss.str()) != std::string::npos);
```
filippobrizzi, line 129:
```c++
ht::internal::log(heph::ERROR, "test another great message", "num", num);
auto log_line = sink_ptr->getLog();
{
    std::stringstream ss;
    ss << "num=" << num;
    EXPECT_TRUE(log_line.find(ss.str()) != std::string::npos);
}

```

modules/telemetry/telemetry/tests/log_tests.cpp
johannes-graeter: aweseome!

filippobrizzi (approved):

modules/telemetry/telemetry/tests/log_tests.cpp
johannes-graeter: this would make the LogTest.sink identical to LogTest.log
The reason behind that test is to test the inner logging mechanics (logging LogEntry) independantly from the varidiac heph::log interface. makes sense or shall i remove the test?

PR #223, opened 2024-12-06
Service client

filippobrizzi:
# Description
Add a client that can be used to make service call to a specific service


modules/ipc/include/hephaestus/ipc/zenoh/service_client.h
```diff
@@ -0,0 +1,46 @@
...
+#include "hephaestus/ipc/zenoh/service.h"
+#include "hephaestus/ipc/zenoh/session.h"
+
+namespace heph::ipc::zenoh {
+
+template <typename RequestT, typename ReplyT>
+class ServiceClient {
+public:
+  ServiceClient(SessionPtr session, TopicConfig topic_config);
+  auto call(const RequestT& request, const std::optional<std::chrono::milliseconds>& timeout = std::nullopt)
```
lhruby, line 22: For robustness reasons, I would not make the timeout optional. The user should always provide a timeout, and reaching the timeout should always notify, log, except.

No timeout is a dead lock.

lhruby (approved):

modules/ipc/include/hephaestus/ipc/zenoh/service_client.h
filippobrizzi: Good point, yes

PR #227, opened 2024-12-11
remove dep to concurrency

johannes-graeter:
# Description

The spinner inside of the consumer module needs logging. 
However up to now we used message_consumer to consume metrics and logs -> cyclic dep
We resovle this by adding the trivial behaviour of the Logger and MetricRecord singletons by hand and remove dep to concurrency.

## Type of change

### DIFF
--- a/modules/telemetry/telemetry/src/log.cpp
+++ b/modules/telemetry/telemetry/src/log.cpp
@@ -58,12 +58,12 @@ Logger::Logger() : entries_{ std::nullopt } {
     while (true) {
       auto message = entries_.waitAndPop();
       if (!message.has_value()) {
-        emptyQueue();
-        return;
+        break;
       }
 
       processEntry(message.value());
     }
+    emptyQueue();
   });
 }
 
--- a/modules/telemetry/telemetry/src/log.cpp
+++ b/modules/telemetry/telemetry/src/log.cpp
@@ -6,7 +6,6 @@
 
 #include <exception>
 #include <future>
-#include <iostream>
 #include <memory>
 #include <optional>
 #include <utility>
@@ -14,6 +13,7 @@
 
 #include <absl/base/thread_annotations.h>
 #include <absl/synchronization/mutex.h>
+#include <fmt/base.h>
 
 #include "hephaestus/containers/blocking_queue.h"
 #include "hephaestus/telemetry/log_sink.h"
--- a/modules/telemetry/telemetry/src/log.cpp
+++ b/modules/telemetry/telemetry/src/log.cpp
@@ -72,7 +72,7 @@ Logger::~Logger() {
     entries_.stop();
     message_process_future_.get();
   } catch (const std::exception& ex) {
-    std::cerr << "While emptying log queue, exception happened: " << ex.what() << "\n";
+    fmt::println(stderr, "While emptying log queue, exception happened: {}", ex.what());
   }
 }
 
--- a/modules/telemetry/telemetry/src/log.cpp
+++ b/modules/telemetry/telemetry/src/log.cpp
@@ -5,6 +5,7 @@
 #include "hephaestus/telemetry/log.h"
 
 #include <exception>
+#include <future>
 #include <iostream>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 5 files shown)


modules/telemetry/telemetry/src/log.cpp
```diff
@@ -28,36 +30,49 @@ class Logger final {
...
+  heph::containers::BlockingQueue<LogEntry> entries_;
+  std::future<void> stop_signal_;
 };
 
-Logger::Logger()
-  : log_entries_consumer_([this](const heph::telemetry::LogEntry& entry) { processLogEntries(entry); },
-                          std::nullopt) {
-  log_entries_consumer_.start();
+Logger::Logger() : entries_{ std::nullopt } {
+  stop_signal_ = std::async(std::launch::async, [this]() {
```
filippobrizzi, line 57: I never know how to properly name future in the case where it is of type void, but maybe I would go for `message_process_future_`

modules/telemetry/telemetry/src/log.cpp
```diff
@@ -28,36 +30,49 @@ class Logger final {
...
+    while (true) {
+      auto message = entries_.waitAndPop();
+      if (!message.has_value()) {
+        emptyQueue();
```
filippobrizzi, line 61: this won't work, as if the message has no value it means the queue is stopped so it will not return any more values. You need to call `emptyQueue` in the destructor before calling queue.stop()

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -121,14 +126,24 @@ void record(const Metric& metric) {
...
 
-MetricRecorder::MetricRecorder()
-  : entries_consumer_([this](const Metric& entry) { processEntries(entry); }, std::nullopt) {
-  entries_consumer_.start();
+MetricRecorder::MetricRecorder() : entries_{ std::nullopt } {
+  stop_signal_ = std::async(std::launch::async, [this]() {
+    while (true) {
+      auto message = entries_.waitAndPop();
+      if (!message.has_value()) {
+        emptyQueue();
```
filippobrizzi, line 134: same as above

modules/telemetry/telemetry/src/log.cpp
johannes-graeter: IMO calling stop() on the blocking queue will do nothing for the impl of log.cpp since it the stop_ flag is only used in waitAndXXX and we are using forcePush and tryPop.
so actually I could delete the stop()...

modules/telemetry/telemetry/src/log.cpp
```diff
@@ -28,36 +30,49 @@ class Logger final {
...
+  message_process_future_ = std::async(std::launch::async, [this]() {
+    while (true) {
+      auto message = entries_.waitAndPop();
+      if (!message.has_value()) {
+        emptyQueue();
+        return;
+      }
+
+      processEntry(message.value());
+    }
```
filippobrizzi, line 66: being very picking here, but I think this would be more explicit:
```suggestion
    while (true) {
      auto message = entries_.waitAndPop();
      if (!message.has_value()) {
        break;
      }

      processEntry(message.value());
    }
    emptyQueue();
```
and same for metrics

modules/telemetry/telemetry/src/log.cpp
```diff
@@ -28,36 +30,49 @@ class Logger final {
...
 }
 
 Logger::~Logger() {
   try {
-    log_entries_consumer_.stop().get();
+    entries_.stop();
+    message_process_future_.get();
   } catch (const std::exception& ex) {
-    std::cerr << "While emptying message consumer, exception happened: " << ex.what() << "\n";
+    std::cerr << "While emptying log queue, exception happened: " << ex.what() << "\n";
```
filippobrizzi, line 75: given that we are updating this, for consistency I think it is better to:
```suggestion
    fmt::println(stderr, "While emptying log queue, exception happened: {}", ex.what());
```

filippobrizzi (approved):
Thanks great work!

PR #231, opened 2024-12-16
Zenoh program option, remove short flag

filippobrizzi:
# Description
To avoid conflicts, we remove all short flag from default zenoh program line argument


rgooj:
lgtm

rgooj (approved):
thanks for the quick fix!

PR #232, opened 2024-12-16
add warn message if no sinks are registered

johannes-graeter:
# Description

when no sink is registered we need some behaviour to cope with it.

We decided to throw an exception instead of defaulting to terminal.

Update:
SInce the exceptoin is hard to add to the async nature of our apporach, we add a print of a warning message to stderr

## Type of change


johannes-graeter:
Also another thing that is ugly is that you need to add a fixture registering the sink to all tests even if they only use the logging as lib... Perhaps defaulting to absl::log with a WARN message might be the better alternative after all...

johannes-graeter:
We decided to not terminate but write a message to cerr in case no sinks are registered.

filippobrizzi (approved):

PR #234, opened 2024-12-18
[StopWatch] add `elapsed` function

filippobrizzi:
# Description
Add `elapsed` function ot `StopWatch`. The function allows to return the time elapsed from first call to `start()`

### DIFF
--- a/modules/utils/include/hephaestus/utils/timing/stop_watch.h
+++ b/modules/utils/include/hephaestus/utils/timing/stop_watch.h
@@ -27,13 +27,17 @@ public:
   [[nodiscard]] auto stop() -> TargetDurationT;
 
   /// \return Currently running lap time:
-  /// - time elapsed from max[most recent start(), most recent lap()] to now.
+  /// - time elapsed from max[most recent start(), most recent lapse()] to now.
   /// - Cast to desired duration.
   /// - Doesn't stop the watch.
   template <typename TargetDurationT = StopWatch::DurationT>
   [[nodiscard]] auto lapse() -> TargetDurationT;
 
-  /// Stop and reset accumlated information.
+  /// \return Elapsed time since the last start() cast to the desired duration.
+  template <typename TargetDurationT = StopWatch::DurationT>
+  [[nodiscard]] auto elapsed() -> TargetDurationT;
+
+  /// Stop and reset accumulated information.
   void reset();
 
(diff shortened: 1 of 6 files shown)


modules/utils/include/hephaestus/utils/timing/stop_watch.h
```diff
@@ -27,13 +27,17 @@ class StopWatch {
...
-  /// - time elapsed from max[most recent start(), most recent lap()] to now.
+  /// - time elapsed from max[most recent start(), most recent lapse()] to now.
   /// - Cast to desired duration.
   /// - Doesn't stop the watch.
   template <typename TargetDurationT = StopWatch::DurationT>
   [[nodiscard]] auto lapse() -> TargetDurationT;
 
-  /// Stop and reset accumlated information.
+  /// \return Elapsed time since the last start() cast to the desired duration.
+  template <typename TargetDurationT = StopWatch::DurationT>
```
lhruby, line 37: I think the clock template parameter should be common to `StopWatch`, and all `StopWatch` member functions should not have a clock template parameter.

As is, you could define one clock for the StopWatch, and another one for `elapsed`, `stop`, ... This seems like a logic error.

modules/utils/include/hephaestus/utils/timing/stop_watch.h
```diff
@@ -27,13 +27,17 @@ class StopWatch {
   [[nodiscard]] auto stop() -> TargetDurationT;
 
   /// \return Currently running lap time:
```
lhruby, line 29:
```suggestion
  /// \return Currently running lap time, measured from the last call to lapse. The first lap is measured from the last start timestamp:
```

lhruby:
LGTM. One suggestion, I would add a simply ascii schematic about what each function does to the class header. That will clarify more than any documentation of the individual functions can
```
|     Lapse      |       Lapse     |
|-----------------|------=-----------|
| accumulatedLapsDuration |
``` 
etc. WDYT?

modules/utils/include/hephaestus/utils/timing/stop_watch.h
filippobrizzi: the clock is indeed the same this is the type of the returned duration you want. in this way you do not have to manually cast between milliseconds and microseconds

PR #239, opened 2025-01-02
[ActionServer] fix: flaky unit test

filippobrizzi:
# Description
Calling stop action server in a loop to make sure we wait enough for the ActionServer to have actually started the stop service

### DIFF
--- a/modules/ipc/tests/action_server_tests.cpp
+++ b/modules/ipc/tests/action_server_tests.cpp
@@ -8,6 +8,7 @@
 #include <thread>
 #include <utility>
 
+#include <absl/log/globals.h>
 #include <fmt/format.h>
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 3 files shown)


modules/ipc/tests/action_server_tests.cpp
```diff
@@ -132,9 +149,22 @@ TEST(ActionServer, ActionServerStopRequest) {
...
+  while (true) {
+    auto success = requestActionServerToStopExecution(*client_session, action_server_data.topic_config);
+    if (success) {
+      break;
+    }
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(4));
+  };
+
+  const auto wait_res = reply_future.wait_for(REPLY_SERVICE_TIMEOUT);
```
johannes-graeter, line 166: so this is the actual fix right? is the REPLY_SERVICE_TIMEOUT only for debugging reasons then?

johannes-graeter (approved):
Nice fix... a comment on the REPLY_SERVICE_TIMEOUT would be good so that its clear that its for debug reasons (if my assumption is correct ;) )

modules/ipc/tests/action_server_tests.cpp
filippobrizzi: the actual fix is the `while(true)`, but the `wait_for` is anyway a good addition

PR #251, opened 2025-01-17
Feat/add test for stringstream

johannes-graeter:
# Description

In order to be able to use the operator<< outside of the module it seems we need to add it to namespace std

## Type of change


filippobrizzi (approved):

johannes-graeter:
Also I sneeked in a downgrade of rules_cc that was requested from bazel

PR #252, opened 2025-01-17
[ActionServer] suboptimally fix bug in call to action server

filippobrizzi:
# Description
See the comment in the code for more information

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
@@ -259,7 +259,21 @@ auto callActionServer(SessionPtr session, const TopicConfig& topic_config, const
         auto client_helper = internal::ClientHelper<RequestT, StatusT, ReplyT>{ session, topic_config,
                                                                                 std::move(status_update_cb) };
 
-        return client_helper.getResponse().get();
+        auto response = client_helper.getResponse().get();
+
+        // TODO(@fbrizzi): fix this properly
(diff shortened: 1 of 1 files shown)


modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
```diff
@@ -259,7 +259,21 @@ auto callActionServer(SessionPtr session, const TopicConfig& topic_config, const
...
+        auto response = client_helper.getResponse().get();
+
+        // TODO(@fbrizzi): fix this properly
+        // This is a temporary fix to avoid the client to return before the server has finished.
+        // If we return immediately what can happen is that we destroy the `response_service_` Service inside
+        // ClientHelper, before it has finished sending the response to the caller (the ActionServer).
+        // This causes Zenoh to panic and the process to segfault.
+        // Ideally, we could have a way for the server to tell us when is idle, but not sure if this is
+        // available.
+        // Also, 100milliseconds it a completely arbitrary value, we will need to increase it if we still have
```
johannes-graeter, line 271:
```suggestion
        // Also, 100milliseconds is a completely arbitrary value, we will need to increase it if we still have
```

johannes-graeter: yeah this is ugly, but I guess if the server has no way of telling us that no jobs are in processing/to-be-processed this might be the only solution for now...

johannes-graeter (approved):
Ugly but necessary given that the server can not tell us that its finished

PR #260, opened 2025-01-23
[Log] add option to set verbosity level to Absl sink

filippobrizzi:
# Description
Absl sink can be instantiated with a verbosity level


modules/telemetry/telemetry/src/log_sinks/absl_sink.cpp
```diff
@@ -7,13 +7,28 @@
...
+      absl::SetGlobalVLogLevel(2);
+      break;
+    case LogLevel::DEBUG:
+      absl::SetGlobalVLogLevel(1);
+      break;
+    case LogLevel::INFO:
+    case LogLevel::WARN:
+    case LogLevel::ERROR:
+    case LogLevel::FATAL:
+      absl::SetGlobalVLogLevel(0);
```
johannes-graeter, line 30: would add a break here, since thats gonna be nasty if someone adds another case.
Wanna add default?

johannes-graeter (approved):

PR #262, opened 2025-01-28
fix cross compiler bug for formatting of chrono

johannes-graeter:
# Description

## Type of change

### DIFF
--- a/modules/examples/examples/CMakeLists.txt
+++ b/modules/examples/examples/CMakeLists.txt
@@ -1,80 +1,80 @@
-#=================================================================================================
+# =================================================================================================
 # Copyright (C) 2023-2024 HEPHAESTUS Contributors
-#=================================================================================================
+# =================================================================================================
 
 define_module_example(
-        NAME zenoh_pub
-        SOURCES zenoh_pub.cpp zenoh_program_options.h
-        PUBLIC_INCLUDE_PATHS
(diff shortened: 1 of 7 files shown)


modules/examples/examples/CMakeLists.txt
```diff
@@ -1,80 +1,80 @@
-#=================================================================================================
+# =================================================================================================
```
johannes-graeter, line 1: seems this wasnt bazel formatted...

modules/utils/include/hephaestus/utils/concepts.h
```diff
@@ -19,6 +19,8 @@
 #include <utility>
 #include <vector>
 
+#include <__chrono/duration.h>
```
filippobrizzi, line 22:
```suggestion
#include <chrono>
```

johannes-graeter: clang doesn't like it... I will deactivate the warning

filippobrizzi (approved):

modules/utils/include/hephaestus/utils/concepts.h
johannes-graeter: interestingly enough, gcc doesn't even publish this header:
 __chrono/duration.h: No such file or directory

PR #263, opened 2025-01-29
Fix cmake, non-exception, clang++18 build and add zstd to mcap

mfehr:
# Description

As part of updating to the newest upstream changes, I enabled unit test and example building in our deployment. This then resulted in the changes below:
 - **fix build for the non-except compile option:** 
    - Address issue where the unit test cannot pass if heph is compiled with `DISABLE_EXCEPTIONS` since it relies on `EXPECT_THROW` in a lot of places -> Introducing `EXPECT_THROW_OR_DEATH` in the tests
    - The actual non-except `CHECK`'s didn't seem to build properly -> changed to `ABSL_CHECK`
 - **`clang++18` build support:**
   The one instance of using std::format did not compile on our system with clang++-18. Since this is only for testing purposes anyways, I dared to propose not compiling that part if the GLIBCXXX is not yet on version 13+. Open to discuss options, maybe there is a better way
 - **Breaking `examples` -> `format` dependency**
   Building the `format` module currently requires cmake 3.23. Because we don't need it and support it, we prefer not building the `format` module. However recent changes added a `format`-example to the `examples` module (which so far are exclusively IPC examples) -> so now `examples` depends on `format`. I moved that example into the `format` module to break that unnecessary dependency.
 - **Add zstd** for mcap compression

## Type of change

- Improvement


cmake/01_version.cmake
```diff
@@ -42,8 +42,8 @@ if(error_version_check)
 endif()
 
 # Extract major, minor and patch versions. NOTE: works only for repo tag in format vX.Y.Z
-if(NOT ${REPO_TAG} MATCHES "^v([0-9]+)\\.([0-9]+)\\.([0-9]+)$")
-  message(FATAL_ERROR "Expected version tag in format v1.2.3, got ${REPO_TAG}")
+if(NOT ${REPO_TAG} MATCHES "^v([0-9]+)\\.([0-9]+)\\.([0-9]+)(-.+)?$")
+  message(FATAL_ERROR "Expected version tag in format v1.2.3 or v1.2.3-*, got ${REPO_TAG}")
```
mfehr, line 46: this was needed, because this version check is tripped up by our v1.2.3-rc1 etc. tags

modules/utils/include/hephaestus/utils/exception.h
```diff
@@ -62,11 +63,19 @@ constexpr void throwExceptionIf(bool condition, const std::string& message,
...
 #endif
 }
 
+#ifdef DISABLE_EXCEPTIONS
+#define EXPECT_THROW_OR_CHECKFAIL(statement, expected_exception, expected_matcher)                           \
+  EXPECT_DEATH(statement, expected_matcher)
+#else
+#define EXPECT_THROW_OR_CHECKFAIL(statement, expected_exception, expected_matcher)                           \
+  EXPECT_THROW(statement, expected_exception)
+#endif
```
filippobrizzi, line 77: Not sure if this can be done, but could we have:
```suggestion
#ifdef DISABLE_EXCEPTIONS
#define EXPECT_THROW(statement, expected_exception, expected_matcher)                           \
  EXPECT_DEATH(statement, expected_matcher)
#endif
```
in this way we do not need to refactor all the tests, but I am not sure if we can redefine a macro.

modules/bag/CMakeLists.txt
```diff
@@ -26,7 +27,13 @@ set(SOURCES
...
 define_module_library(
   NAME bag
-  PUBLIC_LINK_LIBS absl::base hephaestus::containers hephaestus::ipc hephaestus::telemetry mcap
+  PUBLIC_LINK_LIBS
+    absl::base  
+    hephaestus::containers
+    hephaestus::ipc
+    hephaestus::telemetry
+    mcap
+    zstd::libzstd
```
filippobrizzi, line 36: I am happy to add `zstd` support but it will have to be added also to `externals/CMakeLists.txt` with `add_cmake_dependency` and to Bazel.

For bazel you can just add `bazel_dep(name = "zstd", version = "1.5.6")` into `MODULE.bazel` and then link in  `mcap.BUILD`  
 
Let me know if you need help with that.

modules/utils/include/hephaestus/utils/exception.h
mfehr: Will check if we can overlay the gtest EXPECT_THROW with this one.

modules/bag/CMakeLists.txt
mfehr: I honestly didn't realize that zstd isn't present anywhere before upstream, because our cmake wrapper had it as a dependency already in there. It's a an optional dependency of mcap IIRC and there are already todo's in bazel to add it, exactly as you suggested. will check.

filippobrizzi: here I do not use the official Bazel wrap of mcap as it is not great, so please have a look at `mcap.BUILD` you just need to add a dependency to `zstd` and remove `MCAP_COMPRESSION_NO_ZSTD`

modules/utils/include/hephaestus/utils/exception.h
mfehr: overwriting did throw a warning that we do not allow.
changed it to EXPECT_THROW_OR_DEATH to be a bit more consistent with the underlying macro naming.

bazel/foreign_cc/mcap.BUILD
```diff
@@ -22,7 +22,6 @@ cc_library(
     ]),
     defines = [
         "MCAP_COMPRESSION_NO_LZ4",
-        "MCAP_COMPRESSION_NO_ZSTD",
```
filippobrizzi, line 25: can you also add to `deps = [":mcap_src", :"@zstd"]`

mfehr:
@filippobrizzi ptal

PR #273, opened 2025-02-05
Mock clock

filippobrizzi:
# Description
Create a mock clock that allows to mock time in unit tests.

Also, fix bug into liveliness


sithhell (approved):
Looks very nice! I like how the stop watch unit tests was simplified and improved by having this mock.

PR #276, opened 2025-02-06
[Fix] StopWatch: remove clock template

filippobrizzi:
# Description
Pass the clock as an input to the constructor to avoid the templated class


sithhell (approved):
Excellent. This simplifies things significantly.

PR #277, opened 2025-02-06
make log noexcept

johannes-graeter:
# Description

we want to log everywhere also in noexcept regions and in catch blocks of main routines.
Hence make log noexcept by
A. adding try...catch around vector since it might throw std::bad_alloc
B. moving throwing input parameter check of BlockingQueue to create method, constructor still public to go around throwing check

## Type of change

### DIFF
--- a/modules/containers/include/hephaestus/containers/blocking_queue.h
+++ b/modules/containers/include/hephaestus/containers/blocking_queue.h
@@ -25,7 +25,7 @@ concept SimilarTo = std::same_as<std::remove_cvref_t<T>, std::remove_cvref_t<U>>
 template <class T>
 class BlockingQueue {
 public:
-  /// Create a queue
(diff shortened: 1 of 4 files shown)


modules/containers/include/hephaestus/containers/blocking_queue.h
```diff
@@ -25,12 +25,19 @@ concept SimilarTo = std::same_as<std::remove_cvref_t<T>, std::remove_cvref_t<U>>
 template <class T>
 class BlockingQueue {
 public:
-  /// Create a queue
+  /// Create a queue, leave constructor public to go around throwing input parameter checks.
```
filippobrizzi: mmm I think we should remove the constructor, it is a weird pattern like this, nobody would use `create`

filippobrizzi (approved):

PR #283, opened 2025-02-07
[IPC] add type info for subscribers

filippobrizzi:
# Description
For every subscriber we also create a type info service

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/service.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/service.h
@@ -89,6 +89,8 @@ auto callServiceRaw(Session& session, const TopicConfig& topic_config, std::span
                     std::chrono::milliseconds timeout)
     -> std::vector<ServiceResponse<std::vector<std::byte>>>;
 
+[[nodiscard]] auto getEndpointTypeInfoServiceTopic(const std::string& topic) -> std::string;
+
 // -----------------------------------------------------------------------------------------------
 // Implementation
 // -----------------------------------------------------------------------------------------------
--- a/modules/ipc/src/zenoh/dynamic_subscriber.cpp
+++ b/modules/ipc/src/zenoh/dynamic_subscriber.cpp
@@ -72,25 +72,23 @@ void DynamicSubscriber::onPublisher(const EndpointInfo& info) {
 }
 
 void DynamicSubscriber::onPublisherAdded(const EndpointInfo& info) {
-  std::optional<serdes::TypeInfo> optional_type_info;
   if (subscribers_.contains(info.topic)) {
     heph::log(heph::ERROR, "trying to add subscriber for topic but one already exists", "topic", info.topic);
     return;
   }
 
+  auto type_info = topic_db_->getTypeInfo(info.topic);
   if (init_subscriber_cb_) {
-    auto type_info = topic_db_->getTypeInfo(info.topic);
     init_subscriber_cb_(info.topic, type_info);
-    optional_type_info = std::move(type_info);
   }
 
   heph::log(heph::DEBUG, "create subscriber", "topic", info.topic);
   subscribers_[info.topic] = std::make_unique<ipc::zenoh::RawSubscriber>(
       session_, ipc::TopicConfig{ info.topic },
-      [this, optional_type_info = std::move(optional_type_info)](const MessageMetadata& metadata,
-                                                                 std::span<const std::byte> data) {
-        subscriber_cb_(metadata, data, optional_type_info);
-      });
+      [this, type_info](const MessageMetadata& metadata, std::span<const std::byte> data) {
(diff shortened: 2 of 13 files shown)


modules/ipc/src/zenoh/dynamic_subscriber.cpp
```diff
@@ -72,25 +72,23 @@ void DynamicSubscriber::onPublisher(const EndpointInfo& info) {
 }
 
 void DynamicSubscriber::onPublisherAdded(const EndpointInfo& info) {
-  std::optional<serdes::TypeInfo> optional_type_info;
   if (subscribers_.contains(info.topic)) {
     heph::log(heph::ERROR, "trying to add subscriber for topic but one already exists", "topic", info.topic);
     return;
   }
 
+  auto type_info = topic_db_->getTypeInfo(info.topic);
```
mfehr, line 80: I have some changes for this part of the code in my branch as well + in the topic database. If the type retrieval fails, I was just not finishing the onPublisherAdded callback.

mfehr (approved):
lgtm

modules/ipc/include/hephaestus/ipc/zenoh/service.h
```diff
@@ -311,8 +321,10 @@ auto callService(Session& session, const TopicConfig& topic_config, const Reques
   static constexpr auto FIFO_QUEUE_SIZE = 100;
   auto replies = session.zenoh_session.get(keyexpr, "", ::zenoh::channels::FifoChannel(FIFO_QUEUE_SIZE),
                                            std::move(options), &result);
-  throwExceptionIf<FailedZenohOperation>(result != Z_OK,
-                                         fmt::format("Failed to call service on '{}'", topic_config.name));
+  if (result != Z_OK) {
+    heph::log(heph::ERROR, "failed to call service, server error", "topic", topic_config.name);
+    return {};
```
johannes-graeter, line 326: empty vector == failure is quite unintuitive, since this change indicates that the caller has a need to handle this error right? how about absl::StatusOr ? this would even give is the possibility to map the Zenoh errors to sth more detailed than OK or NotOK

modules/ipc/src/zenoh/service.cpp
```diff
@@ -2,4 +2,79 @@
...
+#include "hephaestus/telemetry/log.h"
+
+namespace heph::ipc::zenoh {
+namespace {
+[[nodiscard]] auto
+getServiceCallResponses(const ::zenoh::channels::FifoChannel::HandlerType<::zenoh::Reply>& service_replies)
+    -> std::vector<ServiceResponse<std::vector<std::byte>>> {
+  std::vector<ServiceResponse<std::vector<std::byte>>> reply_messages;
+  for (auto res = service_replies.recv(); std::holds_alternative<::zenoh::Reply>(res);
+       res = service_replies.recv()) {
```
johannes-graeter, line 37: to me its not clear, why you have to receive twice here, an explanatory comment would be great

modules/ipc/src/zenoh/service.cpp
```diff
@@ -2,4 +2,79 @@
...
+
+  const ::zenoh::KeyExpr keyexpr(topic_config.name);
+  ::zenoh::ZResult result{};
+  static constexpr auto FIFO_QUEUE_SIZE = 100;
+  auto replies = session.zenoh_session.get(keyexpr, "", ::zenoh::channels::FifoChannel(FIFO_QUEUE_SIZE),
+                                           std::move(options), &result);
+  if (result != Z_OK) {
+    heph::log(heph::ERROR, "failed to call service, server error", "topic", topic_config.name);
+    return {};
+  }
```
johannes-graeter, line 72: StatusOr?

johannes-graeter (approved):

modules/ipc/src/zenoh/dynamic_subscriber.cpp
filippobrizzi: I think it is easier if you make the PR for that

modules/ipc/include/hephaestus/ipc/zenoh/service.h
filippobrizzi: but the end effect is still the same, no? you make a query and you receive 0 responses. Does it matter why it failed? we had the log in any case.
I would say unless there is a clear case why we need more details on the fail I would keep it simple.
As a general rule, I prefer to do a refactor later than presume something is going to be needed eventually.

modules/ipc/src/zenoh/service.cpp
filippobrizzi: this is a for loop, the first recv is the initialization, the second is the increment

modules/ipc/src/zenoh/service.cpp
filippobrizzi: see above

PR #284, opened 2025-02-07
[IPC] add service type info

filippobrizzi:
# Description

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/service.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/service.h
@@ -37,6 +37,7 @@
 #include "hephaestus/ipc/zenoh/session.h"
 #include "hephaestus/serdes/protobuf/concepts.h"
 #include "hephaestus/serdes/serdes.h"
+#include "hephaestus/serdes/type_info.h"
 #include "hephaestus/telemetry/log.h"
 #include "hephaestus/utils/exception.h"
 
@@ -63,6 +64,7 @@ public:
 
 private:
   void onQuery(const ::zenoh::Query& query);
+  void createTypeInfoService();
 
 private:
   SessionPtr session_;
@@ -73,6 +75,9 @@ private:
   Callback callback_;
   FailureCallback failure_callback_;
   PostReplyCallback post_reply_callback_;
+
+  serdes::ServiceTypeInfo type_info_;
+  std::unique_ptr<Service<std::string, std::string>> type_info_service_;
 };
 
 template <typename ReplyT>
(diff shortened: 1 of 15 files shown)


modules/ipc/src/zenoh/topic_database.cpp
```diff
@@ -64,6 +70,33 @@ auto ZenohTopicDatabase::getTypeInfo(const std::string& topic) -> const serdes::
   return topics_type_db_[topic];
 }
 
+[[nodiscard]] auto ZenohTopicDatabase::getServiceTypeInfo(const std::string& topic)
```
mfehr, line 73: here as well,  I think there shouldn' t be an exception or checkfail if the retrieval fails. I have seen it fail before for publisher types. If we return an std::optional, the recepient can deal with the problem without try/catch. For us it even is not recoverable, because we disable exceptions -> this checkfails.

mfehr (approved):
lgtm

modules/ipc/src/zenoh/service.cpp
```diff
@@ -75,6 +79,15 @@ auto callServiceRaw(Session& session, const TopicConfig& topic_config, std::span
...
 auto getEndpointTypeInfoServiceTopic(const std::string& topic) -> std::string {
-  return fmt::format("type_info/{}", topic);
+  return fmt::format("{}/{}", TOPIC_INFO_SERVICE_TOPIC_PREFIX, topic);
+}
+
+auto isEndpointTypeInfoServiceTopic(const std::string& topic) -> bool {
+  std::vector<std::string> elements = absl::StrSplit(topic, '/');
+  if (elements.empty()) {
+    return false;
+  }
```
johannes-graeter, line 89: you could use std::views::split for this

johannes-graeter (approved):

modules/ipc/include/hephaestus/ipc/zenoh/service.h
```diff
@@ -307,6 +318,21 @@ void Service<RequestT, ReplyT>::onQuery(const ::zenoh::Query& query) {
...
+    return;
+  }
+
+  type_info_service_ = std::make_unique<Service<std::string, std::string>>(
+      session_, TopicConfig{ getEndpointTypeInfoServiceTopic(topic_config_.name) },
+      [this](const std::string&) { return this->type_info_.toJson(); });
+}
+
+// -----------------------------------------------------------------------------------------------
+//
```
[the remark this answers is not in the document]
filippobrizzi: cleaned

modules/ipc/src/zenoh/topic_database.cpp
filippobrizzi: yes, this is a very good point, I will update the behavior for both function in a follow up PR

modules/ipc/src/zenoh/service.cpp
filippobrizzi: nice!

PR #286, opened 2025-02-11
Topic database handle failures

filippobrizzi:
# Description
We need to be able to handle failure in retrieving type info without terminating


mfehr (approved):
lgtm

PR #289, opened 2025-02-13
[IPC/Zenoh] add more options to publisher subscriber and services

filippobrizzi:
# Description
Allow for more granular configuraiton of `Publisher` `Subscriber` and `Service` .


sithhell (approved):
LGTM!

The only (minor?) comment I have is that I disagree wrt backwards compatibility. The interface to create a subscriber/publisher changed. So I guess user code would need to adapt somewhat.

filippobrizzi:
> LGTM!
> 
> The only (minor?) comment I have is that I disagree wrt backwards compatibility. The interface to create a subscriber/publisher changed. So I guess user code would need to adapt somewhat.

You are right it is not strckly backward compatible, adapted the description

mfehr (approved):
lgtm

PR #290, opened 2025-02-13
Make Action server stable

filippobrizzi:
# Description
* Response and status topic are now randomize upon making the call
  * The caller passes the response and status topic to the server, this is needed for the next point.
* The `ClientHelper` is created before making the call the to server to make sure that is already up and running when the action server returns
* Polished the test to have a single session for all the cases.


modules/ipc/include/hephaestus/ipc/zenoh/action_server/types.h
```diff
@@ -5,9 +5,17 @@
...
 
 #include <cstdint>
+#include <string>
 
 namespace heph::ipc::zenoh::action_server {
 
+template <typename RequestT>
+struct Request {
+  RequestT request;
+  std::string response_service_topic_name;
```
sithhell, line 15: wouldn't it be sufficient to just pass in the base topic name and the uid and derive the two topic on top of that instead of sending two strings?

sithhell (changes requested):

PR #296, opened 2025-02-20
Loki sink

filippobrizzi:
# Description
Add log sink for Loki.

At the moment is very simple, and it directly flush every message it receives. 
It can be improved by batching multiple messages and also adding support for the protobuf message.

image


sithhell (approved):
I like it! I just have one nit: how would we test this?

PR #299, opened 2025-03-01
Module - Hephaestus <-> Websocket Bridge 

mfehr:
image

# Description

This module offers a websocket interface to hephaestus, enabling foxglove studio or a similar websocket-capable client to connect to hephaestus communication and:
  - Publish and subscribe to hephaestus topics
  - Trigger hephaestus service calls
  - Monitor the hephaestus node graph (Foxglove has a built in node graph visualizer)

At a high-level, this means for hephaestus:
 - This is the equivalent of https://github.com/foxglove/ros-foxglove-bridge, but for hephaestus
 - We get "RVIZ"-like functionality and some RQT goodies (like visualizing the IPC graph) via foxglove studio
 - Creates a nice interface to connect a something like an Android/Flutter/iOS app to the hephaestus data stream

For more details, check out the README

## Type of change

## Limitations
 - This bridge currently only supports protobuf encoding. This means that some native foxglove studion functions, like calling service calls or some built-in client-side messages are currently not supported because they only work with JSON.
 
## DEPENDENCIES

 - Fork of ws-protocol: https://github.com/olympus-robotics/ws-protocol/pull/1

## Follow Up Work Planned

- [ ] Heph needs properly async service calling  /  a better RPC model
- [ ] Client side requests are processed by a worker pool. By default this was set to 1 thread, likely for simplicity sake. However, this meant that any client-side long-running operation blocked the entire dataflow from the client. Going from 1 to multiple threads, fixes this -> likely we lose client-side message ordering guarantees. Investigate and fix if an issue
- [ ] Actually make the blacklist/whitelist settings do something, currently they are mostly ignored.
- [ ] Test some of the advanced websocket interface functions, if they actually work. If not, do not expose them. (Stuff like TLS and Compression)
- [ ] Add an interface for message transfomers. The idea is that you can register different types of Transfomers to modify the types, add fields to the messages or throttle / combine messages, etc.


modules/websocket_bridge/examples/test_client__services.cpp
```diff
@@ -0,0 +1,260 @@
+#include <algorithm>
```
filippobrizzi, line 1: nit. I would not call this `test_` as it's an example

modules/websocket_bridge/README.md
```diff
@@ -0,0 +1,147 @@
...
+WsBridge --> WsBridgeConfig
+WsBridge --> WsBridgeState
+WsBridge --> IpcGraph
+WsBridge --> IpcInterface
+WsBridge --> Foxglove WebSocket Server
+```
+
+## Test Clients
+
+### Test Client - Services - C++
```
filippobrizzi, line 123: Do I understand correctly that this client is used to test the bridge instead of running foxglove studio? In that case this would be a server, not a client, no? Or am I missing something?

modules/websocket_bridge/include/hephaestus/ipc/ipc_graph.h
```diff
@@ -0,0 +1,158 @@
+//=================================================================================================
```
filippobrizzi, line 1: QQ: this file is going to be moved to the `ipc` module, right?

modules/websocket_bridge/include/hephaestus/ipc/ipc_interface.h
```diff
@@ -0,0 +1,102 @@
...
+      ABSL_EXCLUSIVE_LOCKS_REQUIRED(mutex_sub_);
+
+  // Publishers
+  /////////////
+  mutable absl::Mutex mutex_pub_;
+  std::unordered_map<std::string, std::unique_ptr<ipc::zenoh::RawPublisher>>
+      publishers_ ABSL_GUARDED_BY(mutex_pub_);
+
+  // NOLINTNEXTLINE(readability-identifier-naming)
+  static void callback_PublisherMatchingStatus(const std::string& topic,
```
filippobrizzi, line 88:
```suggestion
  static void publisherMatchingStatusCallback(const std::string& topic,
```

modules/websocket_bridge/include/hephaestus/ipc/ipc_interface.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace heph::ws {
+
+using TopicSubscriberWithTypeCallback = std::function<void(
+    const ipc::zenoh::MessageMetadata&, std::span<const std::byte>, const serdes::TypeInfo&)>;
+
+using RawServiceResponses = std::vector<ipc::zenoh::ServiceResponse<std::vector<std::byte>>>;
+using AsyncServiceResponseCallback = std::function<void(const RawServiceResponses&)>;
+
+class IpcInterface {
```
filippobrizzi, line 35: nit. `IpcEntityManager`

modules/websocket_bridge/include/hephaestus/utils/protobuf_serdes.h
```diff
@@ -0,0 +1,139 @@
+//=================================================================================================
```
filippobrizzi, line 1: this file paht should be
`modules/websocket_bridge/include/hephaestus/websocket_bridge/utils/protobuf_serdes.h`

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
filippobrizzi, line 1: this file should be:
`modules/websocket_bridge/include/hephaestus/websocket_bridge/utils/ws_client.h`

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
```diff
@@ -0,0 +1,60 @@
...
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include <foxglove/websocket/websocket_client.hpp>
+#include <foxglove/websocket/websocket_notls.hpp>
+#include <google/protobuf/message.h>
+#include <hephaestus/utils/ws_protocol.h>
```
filippobrizzi, line 15:
```suggestion
#include "hephaestus/websocket_bridge/utils/ws_protocol.h"
```

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
```diff
@@ -0,0 +1,60 @@
...
+  std::chrono::steady_clock::time_point dispatch_time;
+  std::chrono::steady_clock::time_point response_time;
+
+  std::optional<WsServiceResponse> response;
+  std::string error_message;
+
+  explicit ServiceCallState(uint32_t call_id);
+
+  auto receiveResponse(const WsServiceResponse& service_response, WsAdvertisements& ws_server_ads)
+      -> std::optional<std::unique_ptr<google::protobuf::Message>>;
```
filippobrizzi, line 35: the whole serdes module is supposed to be agnostic from the underlying serialization mechanism, so here we could return just the byte buffer.
Or we want to explicitly state that this bridge works only for protobuf messages?

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
```diff
@@ -0,0 +1,60 @@
...
+#include <foxglove/websocket/websocket_client.hpp>
+#include <foxglove/websocket/websocket_notls.hpp>
+#include <google/protobuf/message.h>
+#include <hephaestus/utils/ws_protocol.h>
+
+namespace heph::ws {
+
+using WsClientNoTls = foxglove::Client<foxglove::WebSocketNoTls>;
+
+struct ServiceCallState {
```
filippobrizzi, line 21: It is a bit of a code smell if a struct has lots of functions. I would either transform this in a class or move the function to free functions, or to another class

modules/websocket_bridge/src/ipc/ipc_graph.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+namespace heph::ws {
+
+IpcGraph::IpcGraph(IpcGraphConfig config) : config_(std::move(config)) {
+}
+
+void IpcGraph::start() {
+  const absl::MutexLock lock(&mutex_);
+
+  fmt::println("[IPC Graph] - Starting...");
```
filippobrizzi, line 34: replce fmt::println with logging

modules/websocket_bridge/tests/test_protobuf_serdes.cpp
```diff
@@ -0,0 +1,117 @@
+//=================================================================================================
```
filippobrizzi, line 1: convention is `protobuf_serdes_tests.cpp`  test files

filippobrizzi (commented):
Great work! look really good, most comments are style one.
In general I do not like shorting names, like `ws` or `WsClient` as the saving is minimal at the cost of making hard to read, but we can find a compromise. I would say either the namespace or the class name should be spelt out

.bazelrc
```diff
@@ -6,5 +6,7 @@ build --repository_cache=~/.cache/bazel/repos/
 build --features=external_include_paths
 build --generate_json_trace_profile --profile command.profile.gz
 
-build:clang-tidy --output_groups=report
+build:clang-tidy --output_groups=report,fixes
 build:clang-tidy --aspects //bazel/clang_tidy:clang_tidy.bzl%clang_tidy_aspect
+build:clang-tidy --action_env=CLANG_TIDY_FLAGS="-fix -fix-errors"
+build:clang-tidy --local_cpu_resources=HOST_CPUS*0.5
```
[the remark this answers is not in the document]
mfehr: yes, clang-tidy repeatedly blew up my 30GB memory...
I didn't intend to commit this though, unless you want to keep it, we can rm.

bazel/foreign_cc/ws_protocol.BUILD
```diff
@@ -0,0 +1,18 @@
...
+    ]),
+    hdrs = glob(["cpp/foxglove-websocket/include/foxglove/websocket/*.hpp"]),
+    includes = ["cpp/foxglove-websocket/include"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "@nlohmann_json//:json",
+        "@boringssl//:ssl",
+        "@websocketpp",
+    ],
+)
```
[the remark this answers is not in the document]
mfehr: curious, I know I can configure VS code to take care of this, but I specifically ran bazel run :format, why doesn't this take care of that?

modules/websocket_bridge/config/ws_bridge-config.yaml
```diff
@@ -0,0 +1,51 @@
+ws_server_config:
+  capabilities:
+    - connectionGraph
+    - clientPublish
+    - services
```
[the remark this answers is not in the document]
mfehr: not my choice, these map directly to the ws-protocol settings in the library I use from foxglove. I could add a translation in between, but not sure if worth it, also introduces another source of error.

modules/websocket_bridge/README.md
mfehr: I (and foxglove) considered the bridge the ws server. So these apps implement alternate ws clients purely in C++ to test the bridge without foxglove studio.

I had to implement them because foxglove studio does not currently support calling service calls via the foxglove studio GUI if the encoding is protobuf. Same for client-side topics.

On the bright side: These serve as half-decent c++ examples for writing additional clients that make use of the bridge, like maybe some development or production tools that you want talking to the robot from the outside without hooking them into the zenoh network.

Does this help?

modules/websocket_bridge/include/hephaestus/ipc/ipc_graph.h
mfehr: I was preparing it for such a move, yes. Will give it a quick try, otherwise it has to be in a follow up PR.

modules/websocket_bridge/include/hephaestus/ipc/ipc_interface.h
mfehr: I quite like the increase in readability this clear  callback_* naming brings. I liked it enough to tell the linter to ignore it. But I guess I'll rename it ;)

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
mfehr: this is a convenience function used in the ws client examples. I could do it in two functions, one returning bytes and the other then the proto message, but the only other encoding option to support that even makes sense for this bridge is json, so I don't think this makes sense and would probably have a lot more places to rework.

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
mfehr: I don't think a class makes sense, I know clang-tidy will then hit me with: no public member variables, which means I add all these accessors, ultimately I end up with an extra 50 lines of code, less readibility for a questionable gain IMO.

I agree on making the functions that actually DO something free functions, because those indeed smell (receive*()). But the convenience accessors which are only there for readability and laziness I would prefer to keep.

modules/websocket_bridge/include/hephaestus/ipc/ipc_interface.h
mfehr: But can we please keep the callback naming in bridge.h / bridge.cpp
It just makes it so much more intutitive not only to recognize callbacks immediately, but see who will be calling them from a architecture PoV.

modules/websocket_bridge/include/hephaestus/ipc/ipc_graph.h
mfehr: Done

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
mfehr: done

modules/websocket_bridge/examples/test_client__services.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // NOLINTNEXTLINE
+  const auto on_open_handler = [&](websocketpp::connection_hdl) { fmt::println("Connected to {}", url); };
+  // NOLINTNEXTLINE
+  const auto on_close_handler = [&](websocketpp::connection_hdl) {
+    fmt::println("Connection closed");
+    g_abort = true;
+  };
+  const auto json_msg_handler = [&](const std::string& json_msg) {
+    handleJsonMessage(json_msg, ws_server_ads, state);
+  };
```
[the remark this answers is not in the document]
filippobrizzi: that could be, but for sure it will give you error if you capture all values

modules/websocket_bridge/include/hephaestus/utils/ws_client.h
filippobrizzi: ok, let's keep it as it is. I think we can safely assume for now we support only proto and then extend it later if needed

filippobrizzi (approved):

modules/ipc/include/hephaestus/ipc/zenoh/ipc_graph.h
```diff
@@ -0,0 +1,157 @@
...
+  IpcGraphConfig config_;
+
+  std::unique_ptr<ipc::zenoh::EndpointDiscovery> discovery_;
+
+  IpcGraphState state_ ABSL_GUARDED_BY(mutex_);
+
+  std::unique_ptr<ipc::ITopicDatabase> topic_db_ ABSL_GUARDED_BY(mutex_);
+
+  TopicRemovalCallback topic_removal_cb_;
+  TopicDiscoveryCallback topic_discovery_cb_;
```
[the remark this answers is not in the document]
mfehr: good catch. I moved that functionality into the config.

modules/ipc/include/hephaestus/ipc/zenoh/ipc_graph.h
```diff
@@ -0,0 +1,157 @@
+//=================================================================================================
```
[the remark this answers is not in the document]
mfehr: Let me see what our AI overlords can do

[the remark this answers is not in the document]
mfehr: done

mfehr:
Thanks for the speedy review @filippobrizzi !

PR #300, opened 2025-03-05
Fix/sensor spinner improve state machine transition logic

lhruby:
# Description
* Change transition logic in the spinner state machine to use states instead of exceptions.
* Move spinner state machine inside a standalone file, as the `Spinner` is independent of the state machine (but not the other way around)

## Type of change

### DIFF
--- a/modules/concurrency/include/hephaestus/concurrency/spinner_state_machine.h
+++ b/modules/concurrency/include/hephaestus/concurrency/spinner_state_machine.h
@@ -9,9 +9,9 @@
 
 namespace heph::concurrency::spinner_state_machine {
 
-enum class State : uint8_t { NOT_INITIALIZED, FAILED, READY_TO_SPIN, SPIN_SUCCESSFUL, EXIT };
-enum class CallbackResult : uint8_t { SUCCESS, FAILURE };
-enum class ExecutionDirective : uint8_t { TRUE, FALSE, REPEAT };
+enum class State : std::uint8_t { NOT_INITIALIZED, FAILED, READY_TO_SPIN, SPIN_SUCCESSFUL, EXIT };
+enum class CallbackResult : std::uint8_t { SUCCESS, FAILURE };
+enum class ExecutionDirective : std::uint8_t { TRUE, FALSE, REPEAT };
 
 using StateMachineCallbackT = std::function<State()>;
 using OperationCallbackT = std::function<CallbackResult()>;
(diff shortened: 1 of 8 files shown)


modules/concurrency/include/hephaestus/concurrency/spinner_state_machine.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include <cstdint>
+#include <functional>
+
+namespace heph::concurrency::spinner_state_machine {
+
+enum class State : std::uint8_t { NOT_INITIALIZED, FAILED, READY_TO_SPIN, SPIN_SUCCESSFUL, EXIT };
+enum class CallbackResult : std::uint8_t { SUCCESS, FAILURE };
```
johannes-graeter, line 13: sad to not have std::expected ...

modules/concurrency/src/spinner_state_machine.cpp
```diff
@@ -0,0 +1,83 @@
...
+  const auto result = params.check();
+  switch (result) {
+    case ExecutionDirective::TRUE:
+      return params.true_state;
+    case ExecutionDirective::FALSE:
+      return params.false_state;
+    case ExecutionDirective::REPEAT:
+      return params.repeat_state;
+  }
+}
```
johannes-graeter, line 46: I don#t really understand the advantage of this pattern.
Probably I am not seeing sth but I would have called a std::function directly and returned the State via a factory

johannes-graeter: I think I see the motivation now, the parameter structs are actually just a way to have named parameter list in c++ right? you could also have the same functions perfectly fine with a normal parameter list but its less nicely readable... i guess since this is a local interface anyways its fine, alhtough it was slightly confusing

johannes-graeter (approved):
like it

filippobrizzi (approved):

modules/concurrency/src/spinner_state_machine.cpp
lhruby: Yes, exactly. I started with functions, but then the state machine logic is somewhat hidden. It's really just to have named parameters.

modules/concurrency/include/hephaestus/concurrency/spinner_state_machine.h
lhruby: Indeed. But we can refactor later.

PR #301, opened 2025-03-07
update chrono duration concept

filippobrizzi:
This type traits doesnt compile with clang 20.1, updating


johannes-graeter (commented):
Great!

PR #302, opened 2025-03-10
feat: add print function to stacktrace

johannes-graeter:
# Description

Add a public interface to print the stacktrace instead of writing to stderr directly.
Prerequisit for Panic specialization.

## Type of change

- [x] If it is a core feature, I have added tests.


johannes-graeter:
How to test the standard behaviour of the stacktrace?

filippobrizzi (approved):

PR #304, opened 2025-03-10
Update clang tidy script

filippobrizzi:
# Description

Update clang_tidy invokation rule to the latest original commit.
Disable `-bugprone-reserved-identifier` as it was the longest check running and doesn't really provide much value.

See Readme to run clang-tidy with profiling enabled


johannes-graeter (approved):
nice even though I kind of have a hard time distinguishingrefactor from functional change...

PR #312, opened 2025-03-13
Make TypeInfo service creation a free function

mfehr:
# Description

Last time we upgraded heph, we ran into this issue here on our side:

```
communication/ipc/ipc/include/ipc/impl/general.h:30:66: error: no type named 'Type' in 'proto_serialization::ProtoAssociation<std::string>'
   30 |   using Type = proto_serialization::ProtoAssociation<ProtoType>::Type;
```
IIRC it was an issue that originated from a unlucky combination of:
 - createTypeInfoService being defined in the header due to it being a member variable of a templated class (Service)
 - the include order of some serdes related header
 We fixed it by moving this function out of Service into a free function, as it doesn't need to be templated at all.

The change should be non-breaking.

As a side note: We also discussed that the proper way would likely be refactoring the Service<T1, T2> into RawService and Service<T1, T2> like it is done with Publisher/Subscriber -> RawPublisher / RawSubscriber. This would have also fixed it since type advertisement could be part of the RawService class. 

## Type of change

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/service.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/service.h
@@ -71,7 +71,6 @@ public:
 
 private:
   void onQuery(const ::zenoh::Query& query);
-  void createTypeInfoService();
 
 private:
   SessionPtr session_;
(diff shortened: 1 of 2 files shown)


filippobrizzi (approved):
nice!

PR #313, opened 2025-03-14
[feat] add formatter for enum

johannes-graeter:
# Description

As per request of @lhruby I added a formatter for enum proxying magic enum.
Just add include it and you should be good to go.
Also works together with generic_formatter as can be seen in the tests.

## Type of change

### DIFF
--- a/modules/format/BUILD
+++ b/modules/format/BUILD
@@ -19,6 +19,7 @@ heph_cc_library(
     deps = [
         "//modules/utils",
         "@fmt",
+        "@magic_enum",
         "@reflect-cpp",
     ],
 )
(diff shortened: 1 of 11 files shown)


johannes-graeter:
Shall we leave it seperate or include it in generic_formatter so that we always have it available when this one is included?

filippobrizzi:
> Shall we leave it seperate or include it in generic_formatter so that we always have it available when this one is included?

@johannes-graeter let's also include it in generic_formatter.h

lhruby (commented):
We have https://github.com/olympus-robotics/hephaestus/blob/main/modules/types/src/dummy_type.cpp
which is using both random and formatting. What about removing the existing solution there, and using your tooling instead? The idea behind types_random is to exactly batch test new functionality for as many types as possible to expose missing features, such as enum serialization.

johannes-graeter:
> > Shall we leave it seperate or include it in generic_formatter so that we always have it available when this one is included?
> 
> @johannes-graeter let's also include it in generic_formatter.h

done

PR #316, opened 2025-03-14
Async Action Call

mfehr:
# Description

Adding an async action call.

## Type of change

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
@@ -108,6 +108,9 @@ private:
 template <typename StatusT>
 using StatusUpdateCallback = std::function<void(const StatusT&)>;
 
+template <typename ReplyT>
+using ResponseCallback = std::function<void(const Response<ReplyT>&)>;
+
 /// Call the action server with the given request.
 /// - The status_update_cb will be called with the status updates from the server.
 ///   - The updates are decided by the user implementation of the action server.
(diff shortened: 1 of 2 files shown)


modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
```diff
@@ -279,4 +292,19 @@ auto callActionServer(SessionPtr session, const TopicConfig& topic_config, const
   });
 }
 
+template <typename RequestT, typename StatusT, typename ReplyT>
+auto callActionServerAsync(SessionPtr session, const TopicConfig& topic_config, const RequestT& request,
```
filippobrizzi, line 296: I am not convinced this is the best way. If the function still returns a std::future it kinds of defeat the purpose of the async as you will still have to call get on the future otherwise you will e hanging forever.

I think here you want to use `std::thread` and then call detach. However, also this solution is not ideal as then you will have no way to know if the thread terminates before closing your app. 
I think the only good way of implementing this is actually to create a client class.

@sithhell do you an idea on what is the best way to implement this function?

sithhell: Ideally, you would implement `callActionServer` in terms of `callActionServerAsync`, not the other way around. The `Async` version will always have the danger of outliving the scope it was called in.
What I am thinking of is something like this:
```
auto callActionServer(...) -> std::future<Response<ReplyT>> {
    std::promise<Response<ReplyT>> promise;
    auto res = promise.get_future();
    callActionServerAsync(..., [promise = std::move(promise)](...) { promise.set_value(...); });
    return res;
}
```

That is: let the callback fulfill the promise of the future. With this, you can implement any asynchronous behavior without additional overhead. While callbacks are hard to program with, they offer the most flexibility.

filippobrizzi: yes but how would you implement the callback based one? as it needs to spawn a thread for the callback and without a client/statefull class it is hard to join the thread

sithhell: ok, I don't see why it needs to spawn a thread? Can we not somehow complete/call the callback through zenoh callbacks?

filippobrizzi: `callActionServer` currently creates a Service to receive the request. It is a free function and the service is owned by the std::async thread.
We would need some entity to own the thread that owns the Service.

Alternativelly, we remove the option to call the actionserver via free function and we create a Client object.

sithhell: Ok, there shouldn't be a thread involved here. The solution would be to capture the state inside a `std::packaged_task` which is invoked (and captured) inside the callback you pass to zenoh.

mfehr: @filippobrizzi @sithhell let's discuss during our next meeting. No urgency to merge this from our side, I just wanted to make an initial version to get the ball rolling. @sithhell  I also remember our consensus to implement callActionServer with Async inside, not the other way around, but I found it not so easy to do that, so I gave the other way around a try first.

mfehr:
stale, we can revisit this when the discussion gains momentum again.

PR #323, opened 2025-04-07
Feat/spinner add telemetry

lhruby:
# Description
Add telemetry to `Spinner`

Example of spinner timings, visualized in Grafana
image

## Type of change
New feature (non-breaking change which adds functionality)

### DIFF
--- a/modules/concurrency/src/spinner.cpp
+++ b/modules/concurrency/src/spinner.cpp
@@ -19,6 +19,7 @@
 #include "hephaestus/telemetry/log.h"
 #include "hephaestus/telemetry/log_sink.h"
 #include "hephaestus/telemetry/metric_record.h"
+#include "hephaestus/utils/timing/stop_watch.h"
 #include "hephaestus/utils/exception.h"
 
 namespace heph::concurrency {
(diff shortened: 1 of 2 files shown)


modules/concurrency/src/spinner.cpp
```diff
@@ -65,21 +72,38 @@ void Spinner::start() {
 void Spinner::spin() {
   // TODO(@brizzi): set thread name
 
-  auto start_timestamp = std::chrono::system_clock::now();
+  const auto timestamp_start = std::chrono::system_clock::now();
+  auto stop_watch = utils::timing::StopWatch();
+  
   while (!stop_requested_.load()) {
     try {
+      stop_watch.start();
```
filippobrizzi, line 80: start has no effect if it was already called, here you want to use `lapse` and ignore the result.
alternatively you can use `start/stop` around the callback

modules/concurrency/src/spinner.cpp
```diff
@@ -65,21 +72,38 @@ void Spinner::start() {
...
+        const auto spin_duration = stop_watch.stop<std::chrono::microseconds>();
+        telemetry::record(telemetry::Metric{
+            .component = component_name_.value(),
+            .tag = "spinner_timestamp",
+            .timestamp = std::chrono::system_clock::now(),
+            .values = { { "callback_duration_microsec",
+                          callback_duration.count() },
+                        { "spin_duration_microsec",
+                          spin_duration.count() } } });
+      }
```
filippobrizzi, line 104: move this before the condition_.wait_until, that should be the last bit of code in the loop

modules/concurrency/src/spinner.cpp
lhruby: check loc 95, I use start/stop here.
lapse has the issue that I want
| callback_duration |              |
| spin_duration                      |

but lapse would yield
| callback_duration | spin_duration - callback_duration |

modules/concurrency/src/spinner.cpp
lhruby: but then I don't get the total spin duration, but only the callback_duration

filippobrizzi (approved):

modules/concurrency/src/spinner.cpp
```diff
@@ -65,21 +71,39 @@ void Spinner::start() {
...
+        telemetry::record(telemetry::Metric{
+            .component = component_name_.value(),
+            .tag = "spinner_timings",
+            .timestamp = std::chrono::system_clock::now(),
+            .values = {
+                { "timestamp_start_spin_microsec", std::chrono::duration_cast<std::chrono::microseconds>(
+                                                       timestamp_start_spin.time_since_epoch())
+                                                       .count() },
+                { "callback_duration_microsec",
+                  std::chrono::duration_cast<std::chrono::microseconds>(callback_duration).count() } } });
```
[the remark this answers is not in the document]
lhruby: Refactored

PR #325, opened 2025-04-08
Update comparison_to_ros.md

filippobrizzi:
Update comparison to ros

### DIFF
--- a/doc/comparison_to_ros.md
+++ b/doc/comparison_to_ros.md
@@ -1,14 +1,27 @@
 # Comparison to ROS
 
-ROS is an excellent framework designed to maximize prototyping speed, flexibility, and code reuse, making it ideal for research labs and early-stage startups. However, building scalable, robust products often demands a different set of priorities: static configuration, determinism, high inspectability, optimized performance, and tailored resource usage.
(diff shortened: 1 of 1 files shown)


doc/comparison_to_ros.md
```diff
@@ -1,12 +1,14 @@
 # Comparison to ROS
 
-* We want to move out of ROS approach to define microservices, where every component is a node and we create a massive data flow graph that gets hard to handle.
-* The benefit of that approach is that it is easy to extend and to compose from existing components, however from our experience robotic systems that are past prototype stage do not have that requirement so the cost of it outweighs the benefit.
-* We have a clear idea of what we want: determinism, easy inspectability and efficiency. This can be done with ROS but it is hard to do.
-* We want a standard message format that is not custom but widely used and highly compatible.
+ROS is an excellent framework designed to maximize prototyping speed, flexibility, and code reuse, making it ideal for research labs and early-stage startups. However, building scalable, robust products often demands a different set of priorities: static configuration, determinism, high inspectability, optimized performance, and tailored resource usage.
```
Fellfalla, line 3: We dont want to support a plug and play ecosystem (same messages for all modules), but philosphy is rather to tailor types to the actual needs to increase quality and support specifics rather than offer unused options. Compatibility between modules is treated as concern of those modules. This targets and favors complete single supplier solutions over solution plumbing and prototyping.

Fellfalla, line 3: Explain why it is hard to do thing right in ROS? -> Offering many ways to achieve the same thing results in inhomogeneous approaches, which ultimately makes inspection and troubleshooting quite difficult.

Fellfalla:
Awesome!

Fellfalla (approved):

PR #329, opened 2025-04-11
[IPC] program options add topic filter to the command line flag; update `EndpointDiscovery` to input a topic filter

filippobrizzi:
# Description
* `EndpointDiscover` now takes in input `TopicFilter` to be more flexible in filtering discovered topic
* `TopicFilter` add option `exclude_prefix` to exclude all topics with a certain prefix
* ipc CLI program options now support topic filter
* `zenoh_topic_list` automatically hides all `topic_info/` topics from the print


sithhell (approved):
LGTM!

PR #333, opened 2025-04-25
[IPC] Fix `RawSubscriber` initialization order

floriantschopp:
Before this change, it was possible that the callback already fired and tried to access the `callback_messages_consumer_` before the constructor finished, resulting in a segfault.


mfehr (approved):

filippobrizzi (approved):
Thanks!

PR #334, opened 2025-04-25
BitFlag: add randomization and reflection

filippobrizzi:
# Description
Add custom random function for bit flag and support for reflection


modules/containers/include/hephaestus/containers/bit_flag.h
```diff
@@ -153,4 +157,26 @@ class BitFlag {
...
+template <containers::UnsignedEnum EnumT>
+[[nodiscard]] auto random(std::mt19937_64& mt) -> containers::BitFlag<EnumT> {
+  containers::BitFlag<EnumT> bit_flag{};
+
+  auto enum_value_count = magic_enum::enum_values<EnumT>().size();
+  auto values = random<std::size_t>(mt, { 0, enum_value_count - 1 });
+  for (std::size_t i = 0; i < values; ++i) {
+    const auto enum_v = random<EnumT>(mt);
+    bit_flag.set(enum_v);
+  }
```
lhruby, line 178: There are N different flags. You can 
* chose each of them and randomly assign true or false, or
* you randomly choose one of the N flags and set it, where you repeat this setting M times (in this case, there will e.g. never be an empty set)
Both approaches are valid and lead to different statistics. I'd naiively go for the first approach, which leads to a normal distribution in the number of flags set, and which accesses the full range of possible values.
 
Please add a comment on what you try to accomplish here.

modules/containers/include/hephaestus/containers/bit_flag.h
```diff
@@ -153,4 +157,26 @@ class BitFlag {
...
+};
+
 }  // namespace heph::containers
+
+namespace heph::random {
+template <containers::UnsignedEnum EnumT>
+[[nodiscard]] auto random(std::mt19937_64& mt) -> containers::BitFlag<EnumT> {
+  containers::BitFlag<EnumT> bit_flag{};
+
+  std::bernoulli_distribution dist;
```
lhruby, line 173: Yes, that is the correct distribution :)

lhruby (approved):

PR #339, opened 2025-05-07
Remove git override; update protobuf

filippobrizzi:
# Description
Using `git_override` is less efficient than archive.

Also, updating protobuf as it is needed by grpc


johannes-graeter (approved):
LGTM

PR #341, opened 2025-05-15
Feat/Serdes unordered map serialization

lhruby:
# Description
Serialize std::unordered_map and extend dummy type.

## Type of change


modules/serdes/include/hephaestus/serdes/protobuf/containers.h
```diff
@@ -26,6 +26,32 @@ namespace heph::serdes::protobuf {
...
+template <Arithmetic T, typename ProtoT>
+void fromProto(const ProtoT& proto_value, T& value) {
+  value = static_cast<T>(proto_value);
+}
+
+template <typename T>
+concept String = std::is_same_v<T, std::string>;
+
+template <String T, typename ProtoT>
+void toProto(ProtoT& proto_value, const T& value) {
```
sithhell, line 46: Why not use `std::string` directly?

modules/serdes/include/hephaestus/serdes/protobuf/containers.h
```diff
@@ -112,4 +138,33 @@ void fromProto(const google::protobuf::RepeatedPtrField<ProtoT>& proto_repeated_
...
+  umap.clear();  // Ensure that the map is empty before adding elements.
+  umap.reserve(static_cast<size_t>(proto_map.size()));
+  for (const auto& proto_pair : proto_map) {
+    std::remove_const_t<K> key;
+    std::remove_const_t<V> value;
+    const auto & [proto_key, proto_value] = proto_pair;
+    fromProto(proto_key, key);
+    fromProto(proto_value, value);
+    umap.emplace(std::move(key), std::move(value));
+  }
```
sithhell, line 167: One potential optimization would be to use the last iterator as a hint.

sithhell (approved):
Just some nitpicking. Looks good otherwise.

modules/serdes/include/hephaestus/serdes/protobuf/containers.h
lhruby: I forgot to remove that part, thanks for the hint. But yeah, using string directly would have been correct.

modules/serdes/include/hephaestus/serdes/protobuf/containers.h
lhruby: Feel free to add a suggestion.

filippobrizzi (approved):
❤️

modules/serdes/include/hephaestus/serdes/protobuf/containers.h
sithhell: I vaguely remembered something about possible optimizations by using [emplace_hint](https://en.cppreference.com/w/cpp/container/unordered_map/emplace_hint).
After looking into it again, it doesn't seem to hold and the performance is similar
This is the test program I used to evaluate: https://quick-bench.com/q/RP7-4a0PpduQgVdC88EY2-mury8
The picture looks similar on my laptop using clang 20.

The article I got the idea from is here: https://fekir.info/post/insert-elements-in-map-efficiently/

PR #345, opened 2025-05-16
add parsing for SecondsT

johannes-graeter:
# Description

add the 'to' function in generic_formatter.h so that time can be read from yaml configs
(Request from @lhruby and @lievenvanderheide )
## Type of change


modules/format/tests/generic_formatter_tests.cpp
```diff
@@ -92,6 +92,46 @@ TEST(GenericFormatterTests, TestFormatKnownObjectWithChronoDuration) {
...
+    const std::string yaml = R"(100)";
+    const auto parsed = rfl::yaml::read<std::chrono::seconds>(yaml);
+    EXPECT_TRUE(parsed.has_value());
+    EXPECT_EQ(*parsed, 100s);
+  }
+
+  // {
+  //   const std::string yaml = R"(100ms)";
+  //   const auto parsed = rfl::yaml::read<std::chrono::milliseconds>(yaml);
+  //   EXPECT_FALSE(parsed.has_value());
```
filippobrizzi, line 131: remove?

filippobrizzi (commented):
This is great! But I would move this code into `containers_serialization` module so that it can be used also for random for example

johannes-graeter:
> This is great! But I would move this code into `containers_serialization` module so that it can be used also for random for example

do you mean containers_reflection?

filippobrizzi:
> > This is great! But I would move this code into `containers_serialization` module so that it can be used also for random for example
> 
> do you mean containers_reflection? Perhaps we should inlcude containers_reflection in generic_formatter, so all reflkecatble types can be formatted?

@johannes-graeter yes, sorry `containers_reflection` and yes then we can add it as a dependency to format

modules/format/include/hephaestus/format/generic_formatter.h
```diff
@@ -60,6 +63,28 @@ struct Reflector<std::chrono::duration<Rep, Period>> {  // NOLINT(misc-include-c
...
     return fmt::format("{:.3f}s", std::chrono::duration_cast<std::chrono::duration<float>>(x).count());
   }
+
+  static auto to(const ReflType& value) -> std::chrono::duration<Rep, Period> {
+    heph::panicIf(value.empty(), "Duration string is empty.");
+    std::string numeric_part;
+    // Check if the string ends with 's'
+    if (value.back() == 's') {
+      // Extract the numeric part by removing the 's' at the end
+      numeric_part = value.substr(0, value.length() - 1);
```
lievenvanderheide, line 73: Is it possible t make numeric_part a std::string_view, in order to avoid the copy and potential allocation?

johannes-graeter: stof needs a string so I might create a stringview and then convert in hte stof call, but with compiler optimizations I would think this should be nearly the same.

modules/containers_reflection/include/hephaestus/containers_reflection/chrono.h
```diff
@@ -0,0 +1,49 @@
...
+    std::string numeric_part;
+    // Check if the string ends with 's'
+    if (value.back() == 's') {
+      // Extract the numeric part by removing the 's' at the end
+      numeric_part = value.substr(0, value.length() - 1);
+    } else {
+      heph::log(heph::ERROR,
+                "Duration string does not end with 's'. Expected format like '123.456s' continuing rest of "
+                "the number",
+                "full string", value);
```
filippobrizzi, line 38: I think here we should panic

johannes-graeter: ok fine with me

modules/format/include/hephaestus/format/generic_formatter.h
lievenvanderheide: If std::stof needs a string, then I wouldn't bother with the string view.

modules/containers_reflection/include/hephaestus/containers_reflection/chrono.h
```diff
@@ -0,0 +1,49 @@
...
+
+namespace rfl {
+
+/// \brief Specialization of the Reflector for chrono based Duration type.
+template <typename Rep, typename Period>
+struct Reflector<std::chrono::duration<Rep, Period>> {  // NOLINT(misc-include-cleaner)
+  using ReflType = std::string;
+
+  static auto from(const std::chrono::duration<Rep, Period>& x) noexcept -> ReflType {
+    return fmt::format("{:.6f}s", std::chrono::duration_cast<std::chrono::duration<float>>(x).count());
```
lhruby, line 24:
```suggestion
    return fmt::format("{:.6f}s", std::chrono::duration_cast<std::chrono::duration<double>>(x).count());
```

lhruby (approved):

modules/format/include/hephaestus/format/generic_formatter.h
johannes-graeter: actually I changed it to use std::from_chars that operates on a view of std::string

PR #350, opened 2025-05-26
Fix/telemetry record performance

sithhell:
# Optimizing `heph::telemetry::record`

- Making `heph::telemetry::record` lazily evaluated
    
    In order to mitigate the overhead of serialization, this patch is
    introducing the capabilities to pass a function to have the metric
    record calculated. This reduces computation time on the caller side and
    moves the costly parts to the service thread.

- Optimizing `BlockingQueue`
    
    With this, we only call notify if there are waiters.

- Adding `UniqueFunction`
    
    This utility can be used in cases where the type erased function does
    not require copying. It only supports moving. Otherwise it is equivalent
    with `std::function`.

## Type of change

### DIFF
--- a/modules/telemetry/telemetry/include/hephaestus/telemetry/metric_record.h
+++ b/modules/telemetry/telemetry/include/hephaestus/telemetry/metric_record.h
@@ -9,9 +9,11 @@
 #include <string>
 #include <string_view>
 #include <unordered_map>
+#include <utility>
 
 #include "hephaestus/serdes/json.h"
 #include "hephaestus/telemetry/metric_sink.h"
+#include "hephaestus/utils/unique_function.h"
 
 namespace heph::telemetry {
 namespace internal {
@@ -28,7 +30,8 @@ void registerMetricSink(std::unique_ptr<IMetricSink> sink);
 /// The metric is forwarded to all registered sinks.
 /// Sinks process the metric in a dedicated thread, this means that this function is non-blocking and
 /// deterministic.
-void record(const Metric& metric);
+void record(Metric metric);
+void record(heph::UniqueFunction<Metric()> metric);
 
 /// @brief Record a user defined metric.
 /// NOTE: the data needs to be serializable to JSON.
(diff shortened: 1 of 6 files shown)


modules/telemetry/telemetry/include/hephaestus/telemetry/metric_record.h
```diff
@@ -39,17 +42,21 @@ void record(const Metric& metric);
...
-  const auto json = serdes::serializeToJSON(data);
-  const Metric metric{
-    .component = component,
-    .tag = tag,
-    .timestamp = timestamp,
-    .values = internal::jsonToValuesMap(json),
-  };
-
-  record(metric);
+void record(std::string component, std::string tag, DataT&& data) {
```
johannes-graeter, line 45: is the change of interface necessary here? i don't see why the timestamp interface needs to be removed

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -115,16 +116,20 @@ class MetricRecorder {
...
   std::future<void> message_process_future_;
 };
 
 void registerMetricSink(std::unique_ptr<IMetricSink> sink) {
   MetricRecorder::registerSink(std::move(sink));
 }
 
-void record(const Metric& metric) {
-  MetricRecorder::record(metric);
+void record(heph::UniqueFunction<Metric()> metric) {
```
johannes-graeter, line 127: since its unique isn't it gonnda be a && always? in that case you can also make the interface narrower directly

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -134,8 +139,7 @@ MetricRecorder::MetricRecorder() : entries_{ std::nullopt } {
       if (!message.has_value()) {
         break;
       }
-
-      processEntry(message.value());
+      processEntry((*message)());
```
johannes-graeter, line 142: @filippobrizzi  do we care if we use * or .value() ?

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -161,9 +165,9 @@ void MetricRecorder::registerSink(std::unique_ptr<IMetricSink> sink) {
   telemetry.sinks_.push_back(std::move(sink));
 }
 
-void MetricRecorder::record(const Metric& metric) {
+void MetricRecorder::record(heph::UniqueFunction<Metric()> metric) {
```
johannes-graeter, line 168: same as above

modules/telemetry/telemetry/src/metric_record.cpp
```diff
@@ -180,7 +184,7 @@ void MetricRecorder::emptyQueue() {
       return;
     }
 
-    processEntry(message.value());
+    processEntry(message.value()());
```
johannes-graeter, line 187: above you used * i don't care which one but be consistent :)

modules/utils/include/hephaestus/utils/unique_function.h
```diff
@@ -0,0 +1,240 @@
...
+
+#include <concepts>
+#include <cstddef>
+#include <functional>
+#include <type_traits>
+
+namespace heph {
+/// The `UniqueFunction` is very similar to `std::function` with the
+/// only difference that it is move only.
+template <typename Sig>
```
johannes-graeter, line 15: I guess in c++23 we can replace this by std::move_only_function?

@filippobrizzi when are we finally able to migrate hephaestus to c++23 keeping it at c++20 starts to suck pretty bad

johannes-graeter: @sithhell if std::move_only_function is a dropin replacement for this, add a comment to remove this when we update

johannes-graeter (approved):
nice work!

some comments but overall good enough to merge.
Could have been 3 MRs though ;)

modules/utils/include/hephaestus/utils/unique_function.h
sithhell: It could be a drop in replacement indeed. Not sure when `std::move_only_function` will land in libc++ though.

modules/telemetry/telemetry/src/metric_record.cpp
sithhell: good point.

PR #354, opened 2025-06-04
Various Conduit Improvements

sithhell:
# Description

- Node triggers are now started in parallel with the periodic trigger
- Adding Intrusive FIFO queue to avoid memory allocations:
  - Used it inside the scheduler and enqueuing input awaiters
- Fixing scheduling drift for periodic nodes
- Allow setting inputs from different threads
- Making `AccumulatedInput` actually work as trigger
- Adding scheduleAt to simplify periodic node scheduling and directly schedule tasks with
  a deadline in the past

## Type of change


modules/containers/include/hephaestus/containers/intrusive_fifo_queue.h
```diff
@@ -0,0 +1,93 @@
...
+struct IntrusiveFifoQueueAccess {
+  template <typename T>
+  static auto next(T* t) -> T*& {
+    if constexpr (requires(T* node) { node->next; }) {
+      return t->next;
+    } else if constexpr (requires(T* node) { node->next_; }) {
+      return t->next_;
+    } else {
+      static_assert(false, "Don't know how to access next pointer");
+    }
```
filippobrizzi, line 19: why not having this function inside `IntrusiveFifoQueue` ?

filippobrizzi (approved):

PR #356, opened 2025-06-10
[Conduit] add node abstraction to connect with zenoh topics

filippobrizzi:
# Description
as per title


modules/conduit/include/hephaestus/conduit/zenoh_nodes.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace heph::conduit {
+
+template <typename InputT>
+class ZenohSubscriberNode {
+public:
+  ZenohSubscriberNode(ipc::zenoh::SessionPtr session, ipc::TopicConfig topic_config, InputT& input)
+    : input_(&input)
+    , subscriber_(std::move(session), std::move(topic_config),
+                  [this](const auto&, const auto& msg) { input_->setValue(*msg); }) {
```
sithhell, line 27: `setValue` might return `InputState::OVERFLOW`. I guess you need to do something like:
```
auto backoff = std::chrono::microseconds(1);
while (input->setValue(*msg) != InputState::OK) {
  std::this_thread::sleep_for(backoff);
  backoff = backoff * 2;
}
```

PR #358, opened 2025-06-12
 Adding basic async IO functionality

sithhell:
# Description
    
 - Adding socket and endpoint class for wrapping POSIX API
 - Using sender/receiver for accept/recv/send operations

## Type of change


modules/net/include/hephaestus/net/recv.h
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  auto operator()(Socket const& socket, std::span<std::byte> buffer) const
+      -> stdexec::__binder_back<RecvT, Socket const*, std::span<std::byte>> {
+    return { { &socket, buffer }, {}, {} };
+  }
+};
+
+// NOLINTNEXTLINE (readability-identifier-naming)
+static inline RecvT<false> recv{};
```
filippobrizzi, line 48: consider using normal functions

modules/net/include/hephaestus/net/send.h
```diff
@@ -0,0 +1,132 @@
...
+  }
+};
+
+// NOLINTNEXTLINE (readability-identifier-naming)
+static inline SendT<false> send{};
+// NOLINTNEXTLINE (readability-identifier-naming)
+static inline SendT<true> sendAll{};
+
+template <bool SendAll, typename Receiver>
+struct SendOperation {
```
filippobrizzi, line 53: move this into `internal` namespace

modules/net/include/hephaestus/net/recv.h
sithhell: unfortunately not possible due to the partial application of the function and the desired `operator|` overload.

modules/net/examples/ping_pong.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "hephaestus/net/acceptor.h"
+#include "hephaestus/net/recv.h"
+#include "hephaestus/net/send.h"
+#include "hephaestus/net/socket.h"
+#include "hephaestus/telemetry/log.h"
+#include "hephaestus/telemetry/log_sinks/absl_sink.h"
+
+namespace {
+inline constexpr unsigned PAGE_SIZE = 4096;
+// NOLINTNEXTLINE (readability-static-accessed-through-instance)
```
filippobrizzi, line 32: why is this needed?

filippobrizzi (approved):

modules/net/examples/ping_pong.cpp
sithhell: There seems to be a issue within the task implementation which access a static variable through an object and clang-tidy starts to complain.

PR #365, opened 2025-06-24
Feat/refactor panic

johannes-graeter:
# Description

For panicIf, we do not want to pay runtime cost in case the condition is false.
* Hence we introduce lazily formatted panicIf
* for consistency we change also the interface of panic
* as a positive side effect this hides fmt from the panic interface having an easier time converting to std::format at some point
* makes the panic interface less verbose

Note:
As a price we need to evaluate the string at runtime and give up the compile time checks of std::format.
Also there is a small runtime overhead caused by the impl of format (type erasure performed there) that should be negligeable.
Alternatively we could use std::format and use a template string. however the interface seems not nice:
```
panicIf<"I am panicking {}">(condition,  "just because");
```

## Type of change
- Breaking change -> interface of panic and panicIf changes

### DIFF
--- a/modules/concurrency/src/io_ring/timer.cpp
+++ b/modules/concurrency/src/io_ring/timer.cpp
@@ -9,7 +9,6 @@
 #include <limits>
 #include <system_error>
 
-#include <fmt/format.h>
 #include <liburing.h>  // NOLINT(misc-include-cleaner)
 #include <liburing/io_uring.h>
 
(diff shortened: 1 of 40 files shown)


modules/concurrency/src/io_ring/timer.cpp
```diff
@@ -39,8 +38,7 @@ void Timer::Operation::prepare(::io_uring_sqe* sqe) const {
 
 void Timer::Operation::handleCompletion(::io_uring_cqe* cqe) const {
   if (cqe->res < 0 && cqe->res != -ETIME) {
-    heph::panic(
-        fmt::format("timer failed: {}", std::error_code(-cqe->res, std::system_category()).message()));
+    heph::panic("timer failed: {}", std::error_code(-cqe->res, std::system_category()).message());
```
filippobrizzi, line 41: while you are at it, would you mind removing all the `heph::` they are not needed given that we are inside that namespace

filippobrizzi (approved):

PR #367, opened 2025-06-25
Websocket Bridge & Liveliness Fixes

mfehr:
# Description

This is a collection of fixes I would like to push upstream resulting from more in-depth integration testing of the websocket bridge and the underlying raw/dynamic pub/sub/service call discovery.

General (need upstream porting)
 - dynamic subscriber - advertised type service call of that subscriber was broken
  - fix accidental type service recursion (i.e. type services of raw subs/pubs, spawned their own type services again)
 - enable liveliness token `history` subscriber option -> Should (**but currently doesn't**) allow for late-joiners to discovery all topics/services
 - websocket bridge
    - websocket bridge didn't forward empty (but valid) service responses
    - generally lower default verbosity of websocket bridge and add option for verbose output
    - actually implement whitelisting/blacklisting of topics/services

## Type of change

### DIFF
--- a/modules/ipc/src/zenoh/liveliness.cpp
+++ b/modules/ipc/src/zenoh/liveliness.cpp
@@ -172,20 +172,7 @@ EndpointDiscovery::EndpointDiscovery(SessionPtr session, TopicFilter topic_filte
   , infos_consumer_(std::make_unique<concurrency::MessageQueueConsumer<EndpointInfo>>(
         [this](const EndpointInfo& info) { callback_(info); }, DEFAULT_CACHE_RESERVES)) {
   infos_consumer_->start();
-  // NOTE: the liveliness token subscriber is called only when the status of the publisher changes.
-  // This means that we won't get the list of publisher that are already running.
-  // To do that we need to query the list of publisher beforehand.
-  auto publishers_info = getListOfEndpoints(*session_, topic_filter_);
-  for (const auto& info : publishers_info) {
-    infos_consumer_->queue().forcePush(info);
-  }
 
-  // Here we create the subscriber for the liveliness tokens.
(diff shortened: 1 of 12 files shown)


modules/websocket_bridge/src/websocket_bridge/utils/ws_protocol.cpp
```diff
@@ -27,10 +27,6 @@ auto convertIpcRawServiceResponseToWsServiceResponse(
     WsServiceId service_id, WsServiceCallId call_id,
     const ipc::zenoh::ServiceResponse<std::vector<std::byte>>& raw_response, WsServiceResponse& ws_response)
     -> bool {
-  if (raw_response.value.empty()) {
-    return false;
-  }
-
```
mfehr, line 33: note that when a service call returns an empty proto object or a proto object has all optional fields unset, I assumed that even then the byte array would never be empty, but turns out it is. So this is a valid response.

modules/ipc/src/zenoh/liveliness.cpp
```diff
@@ -208,9 +210,12 @@ void EndpointDiscovery::createLivelinessSubscriber() {
     }
   };
 
+  ::zenoh::ZResult result{};
   liveliness_subscriber_ =
       std::make_unique<::zenoh::Subscriber<void>>(session_->zenoh_session.liveliness_declare_subscriber(
-          keyexpr, std::move(liveliness_callback), ::zenoh::closures::none));
+          keyexpr, std::move(liveliness_callback), ::zenoh::closures::none, { .history = true }, &result));
```
filippobrizzi, line 216: This works great! while you are at it, would you mind removing the code in the constructor that calls `getListOfEndpoints`? As it is no longer needed with this change

PR #371, opened 2025-06-27
 Cleaning up net module

sithhell:
# Description
    
- Instead of having enums as ctor parameters, use create factory
 functions
- Pass on the context directly to the socket to avoid needing it to pass
 to each send/recv/accept.

## Type of change

### DIFF
--- a/modules/net/examples/bt_server.cpp
+++ b/modules/net/examples/bt_server.cpp
@@ -25,6 +25,7 @@
 #include "hephaestus/net/socket.h"
 #include "hephaestus/telemetry/log.h"
 #include "hephaestus/telemetry/log_sinks/absl_sink.h"
+#include "hephaestus/utils/signal_handler.h"
 
 namespace {
 inline constexpr std::size_t PACKET_SIZE = 65535;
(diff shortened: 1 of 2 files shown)


modules/net/examples/bt_server.cpp
```diff
@@ -26,24 +25,21 @@
...
 namespace {
 inline constexpr std::size_t PACKET_SIZE = 65535;
 inline constexpr double KB = 1024.;
-// NOLINTNEXTLINE (readability-static-accessed-through-instance)
-auto pong(heph::concurrency::ContextScheduler scheduler, heph::net::Socket socket) -> exec::task<void> {
+auto pong(heph::net::Socket socket) -> exec::task<void> {
   std::array<char, PACKET_SIZE> buffer{};
 
-  while (!heph::utils::TerminationBlocker::stopRequested()) {
+  while (true) {
```
filippobrizzi, line 35: how do you terminate it now?

modules/net/src/endpoint.cpp
```diff
@@ -124,20 +99,21 @@ auto getPortBt(const std::vector<std::byte>& address) -> std::uint16_t {
...
+  switch (type()) {
+    case heph::net::EndpointType::IPV4:
       return getPortIpV4(address_);
-    case AF_INET6:
+    case heph::net::EndpointType::IPV6:
       return getPortIpV6(address_);
-    case AF_BLUETOOTH:
+    case heph::net::EndpointType::BT:
       return getPortBt(address_);
     default:
```
filippobrizzi, line 109: nit. it is better not to have default so if you add a new case the compiler will complain if you forget to address it somewhere

filippobrizzi (approved):

modules/net/src/endpoint.cpp
sithhell: good point.

PR #383, opened 2025-07-14
FIx/Reset action server status

mgeisslinger:
# Description

The pollable action server publishes feedback continuously instead of on every change.
This MR fixes this by resetting the optional.

## Type of change


modules/ipc/include/hephaestus/ipc/zenoh/action_server/pollable_action_server.h
```diff
@@ -133,7 +133,8 @@ PollableActionServer<RequestT, StatusT, ReplyT>::PollableActionServer(SessionPtr
             }
 
             if (status_.has_value()) {
-              status_publisher.publish(*std::move(status_));
+              status_publisher.publish(std::move(*status_));
```
filippobrizzi, line 136: I would remove std::move alltogether as the publish function anyway takes a reference

PR #388, opened 2025-07-15
Feat/add leaner accumulated input interface

johannes-graeter:
# Description

make interface of accumulated input leaner with an alias.
Concept for it is deliberately not added to utils/concepts.h to avoid the dependency.

This is an API break. AFAIK this feature is not used yet so this is a good point to change the API.

## Type of change


modules/conduit/include/hephaestus/conduit/accumulated_input.h
```diff
@@ -46,10 +47,17 @@ class AccumulatedInput
...
   F f_;
   R initial_value_;
 };
+
+template <typename R>
+concept WithValueType = requires { typename R::value_type; };
+
+template <WithValueType R, typename InputPolicy = InputPolicy<>>
+using AccumulatedInput =
+    AccumulatedInputBase<typename R::value_type, R, std::function<R(typename R::value_type, R)>, InputPolicy>;
```
sithhell, line 62: Can we have a generic alias without the `WithValueType` constrained as well?

filippobrizzi (approved):
LGTM!

PR #392, opened 2025-07-21
Fix: Improve FIFO queue, stabilize timer

sithhell:
# Description

- FIFO is now a doubly linked circular list (avoids linear time complexity for task removal)
- Minor adjustements for timer

## Type of change

### DIFF
--- a/modules/concurrency/src/io_ring/timer.cpp
+++ b/modules/concurrency/src/io_ring/timer.cpp
@@ -45,7 +45,15 @@ void Timer::Operation::handleCompletion(::io_uring_cqe* cqe) const {
   timer->tick();
 }
 
-void Timer::Operation::handleStopped() {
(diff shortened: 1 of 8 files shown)


modules/concurrency/src/io_ring/timer.cpp
```diff
@@ -147,13 +148,13 @@ auto Timer::tickSimulated(bool advance) -> bool {
   }
 
   if (advance) {
-    const auto& top = tasks_.front();
+    auto top = tasks_.front();
+    std::ranges::pop_heap(tasks_, std::greater<>{});
```
johannes-graeter, line 152: interesting. do you need to protect pop_heap with std::ranges::is_heap or is that guaranteed?

johannes-graeter (approved):
nice just 1 question that you is more out of interest

PR #405, opened 2025-09-24
Fix/Action Server: Lifecycle of stop service

mgeisslinger:
# Description
The ActionServer::execute method created its stop_service as a local variable on the stack. When the execute method returned, the Service object was destroyed, but its callback remained registered within Zenoh. A subsequent call to the stop service would invoke a dangling function pointer, resulting in a stack-use-after-return in address sanitizer.

The stop_service has been promoted to a class member of ActionServer, owned by a std::unique_ptr. It is now created in the ActionServer's constructor, tying its lifetime to the ActionServer itself and ensuring its callback is always valid.

## Type of change

### DIFF
--- a/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
+++ b/modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
@@ -169,12 +169,11 @@ ActionServer<RequestT, StatusT, ReplyT>::ActionServer(SessionPtr session, TopicC
   , request_consumer_([this](const Request<RequestT>& request) { return execute(request); }, std::nullopt)
   , stop_service_(std::make_unique<Service<std::string, std::string>>(
         session_, internal::getStopServiceTopic(topic_config_),
-        [this](const std::string&) {  // Main callback using 'this'
+        [this](const std::string&) {
           this->stop_requested_ = true;
           return "stopped";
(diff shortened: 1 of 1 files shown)


modules/ipc/include/hephaestus/ipc/zenoh/action_server/action_server.h
```diff
@@ -163,11 +166,21 @@ ActionServer<RequestT, StatusT, ReplyT>::ActionServer(SessionPtr session, TopicC
                  .status = serdes::getSerializedTypeInfo<StatusT>() })
   , type_info_service_(createTypeInfoService(
         session_, topic_config_, [this](const std::string&) { return this->type_info_.toJson(); }))
-  , request_consumer_([this](const Request<RequestT>& request) { return execute(request); }, std::nullopt) {
+  , request_consumer_([this](const Request<RequestT>& request) { return execute(request); }, std::nullopt)
+  , stop_service_(std::make_unique<Service<std::string, std::string>>(
+        session_, internal::getStopServiceTopic(topic_config_),
+        [this](const std::string&) {
+          this->stop_requested_ = true;
```
filippobrizzi, line 173: we need to be carefull, here if we call stop and the service is not running, but then we start running the stop_requested_ would be true and we will never execute the request

PR #459, opened 2025-12-03
add bazel support to gcc and upgrade to C++23

filippobrizzi:
# Description
Add GCC toolchain and upgrade to GCC 13 and C++23.

Issue: GCC uses `iquote` instead of `isystem` for abseil dependency and the compiler doesn't find it.
It happens only for abseil and not for fmt or others. No idea what is happening.

To reproduce: `bazel build //modules/utils/... --config gcc`


johannes-graeter:
why not go to gcc15 and c++26 preview?

